PR #4, opened 2024-01-03
Further Microblog/Entry Support

olorin99:
Added
- Viewing posts in magazine/user feeds.
- Edit posts and post comments.
- Edit entries entry comments .
- Delete posts and post comments.
- Delete entries and entry comments.
- Edit/Delete buttons are greyed out on feed page.
- Edit/Delete buttons are greyed out if logged in user is not op.

### DIFF
--- a/lib/src/api/comments.dart
+++ b/lib/src/api/comments.dart
@@ -175,3 +175,39 @@ Future<Comment> postComment(
 
   return Comment.fromJson(jsonDecode(response.body) as Map<String, dynamic>);
 }
+
+Future<Comment> editComment(
+    http.Client client,
+    String instanceHost,
+    int commentId,
+    String body,
+    String lang,
+    bool? isAdult
+    ) async {
+  final response = await client.put(Uri.https(
+      instanceHost,
+      '/api/comments/$commentId'
+  ),
+      body: jsonEncode({
+        'body': body,
+        'lang': lang,
+        'isAdult': isAdult ?? false
+      }));
+
+  httpErrorHandler(response, message: "Failed to edit comment");
+
+  return Comment.fromJson(jsonDecode(response.body) as Map<String, dynamic>);
+}
+
+Future<void> deleteComment(
+    http.Client client,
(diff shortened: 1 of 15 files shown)


jwr1 (changes requested):
LGTM, I just have one small change. Could you have it so when editing entries/posts/comments that they get prefilled with the original content? This could be an `initEdit` function parameter in the `ActionBar` widget that fetches the original content and sets the text controller with it.

jwr1 (changes requested):
One other thing I noticed is that the edit/delete menu still shows up and is clickable even when the thread/post/comment was made by someone else.

jwr1 (changes requested):
Also (hopefully this is the last thing), when the edit and delete buttons shouldn't be available, like on the feed view or when logged out, the delete button is greyed out correctly, but the edit button is not.

image

olorin99:
> One other thing I noticed is that the edit/delete menu still shows up and is clickable even when the thread/post/comment was made by someone else.

In order to check if logged in user is the OP, is the logged in userID stored somewhere?
I can see "_selectedAccount" in SettingsController. Would this be the best variable to use for comparisons?

jwr1:
> > One other thing I noticed is that the edit/delete menu still shows up and is clickable even when the thread/post/comment was made by someone else.
> 
> In order to check if logged in user is the OP, is the logged in userID stored somewhere? I can see "_selectedAccount" in SettingsController. Would this be the best variable to use for comparisons?

Yes, that would work.

jwr1:
`selectedAccount` isn't the user id though, it's the username (like @hello@kbin.earth, instead of 53198).

jwr1:
It looks like the username, from the API, for non federated users is just the username itself (no at symbols or hosts), so you could just check that the user.username field in content matches the part before the `@` in `selectedAccount`.

olorin99:
@jwr1 Should be good to go now.

jwr1 (approved):
Looks good, thanks!

PR #8, opened 2024-01-20
Add OP marker in comments.

olorin99:
Adds a OP marker on comments made by thread/post op. Also includes bug fix where edit/delete comments weren't greyed out.


jwr1 (approved):
LGTM, thanks!

PR #11, opened 2024-01-29
Themes

olorin99:
Adds ability to use dynamic colours
- On android used wallpaper colour
- On ios uses app accent colour
- On linux uses GTK+ theme colour

Adds themes and option to change current theme.
Current themes:
- Default theme
- Dynamic
- Catppuccin
- Blue
- Purple
- Red
- Amber

Themes can be added or changed by defining light and dark colour schemes in utils/themes.dart.


jwr1 (changes requested):
Looks good so far; I just have one change request. Instead of using an integer for saving theme accent preference, could you use a string with the theme's name? That way, if we reorder, remove, or add more themes, the current user's preference won't be messed up.

PR #14, opened 2024-02-06
Search

olorin99:
Adds general search with screen accessed from explore screen.
Search feed shows Entries/Posts/Comments/User/Magazines mixed together.
Search for users using @user@instance and magazines using @magazine@instance.
Should federate searched users and magazines if not federated already.
Also add option for comments which goes to the original post.


jwr1 (approved):
Looks great! I just made a few small changes, like changing the ternary to a switch and a some UI tweaks, but I think it's good to merge now. Thanks!

PR #16, opened 2024-02-12
Some updates to profile page

olorin99:
Adds profile cover image.
Adds optional border for avatars.
Updates profile page to show username with and without instance info. 
Clicking on username with instance info (e.g. "@olorin99@kbin.earth") copies username to clipboard.


jwr1 (approved):
Looks great, thanks!

PR #18, opened 2024-02-14
Adds highlighting for user mentions.

olorin99:
Replaces @mentions in markdown with a inkwell that routes to the users profile when clicked.

### DIFF
--- a/lib/src/api/users.dart
+++ b/lib/src/api/users.dart
@@ -39,6 +39,20 @@ Future<DetailedUserModel> fetchUser(
       jsonDecode(response.body) as Map<String, dynamic>);
 }
 
+Future<DetailedUserModel> fetchUserByName(
+    http.Client client,
+    String instanceHost,
+    String username,
+    ) async {
+  final response =
+  await client.get(Uri.https(instanceHost, '/api/users/name/$username'));
+
+  httpErrorHandler(response, message: 'Failed to load user');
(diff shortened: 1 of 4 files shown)


olorin99:
Lol didn't pull before pushing this.

jwr1:
The functionality of user/magazine mentions should be done, but I'm still working on the visual indicator. At the moment, this is how it looks: 

image

jwr1:
It should be done now, if you want to try it out or make suggestions.

olorin99:
Looks great. One thing I noted is that apparently magazine mentions can also be @magazine. The main use case seems to be [mastodon users posting in lemmy communities](https://kbin.social/m/showerthoughts@lemmy.world/p/4515910/At-some-point-in-your-childhood-you-and-your-friends)

However it seems to conflict with user mentions like at https://kbin.social/m/kbinMeta/t/828882/So-what-s-going-on where the @tech links to a user instead of the magazine. So it is probably better not to handle that case.

jwr1:
I hadn't thought of that. Microblogging is half of kbin, but I don't think the use cases outway the weird problems it could cause. We can easily add it later on though, if we think it'd be useful.

olorin99:
Agreed. In the few cases I could find the link to the magazine is right there anyway next to the username so there's little to no impact on usability.

PR #19, opened 2024-02-15
Add support for editing users profile.

olorin99:
Adds editing about section, avatar image and cover image.

### DIFF
--- a/lib/src/screens/explore/user_screen.dart
+++ b/lib/src/screens/explore/user_screen.dart
@@ -302,6 +302,7 @@ class _UserScreenState extends State<UserScreen> {
                               : TextEditor(
                                 _aboutTextController!,
                                 label: "About",
+                                isMarkdown: true,
                               )
                         ),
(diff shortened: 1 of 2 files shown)


lib/src/screens/explore/user_screen.dart
```diff
@@ -218,14 +291,49 @@ class _UserScreenState extends State<UserScreen> {
...
-                        )
+                          child: _aboutTextController == null
+                              ? Markdown(
+                                _data!.about!,
+                                getNameHost(context, _data!.username),
+                              )
+                              : TextEditor(
+                                _aboutTextController!,
+                                label: "About",
+                              )
```
jwr1: Looks good. The only thing needed is for the "About" text editor to have `isMarkdown` set to `true`. That way, multi-line input is enabled.

jwr1 (changes requested):

jwr1 (approved):
Thanks!

PR #22, opened 2024-02-21
Couple of fixes for microblogs and mentions.

olorin99:
Updated microblog pagekey to match entries for feed screen.
For the mention cache, the comparison to the originInstance broke mentions for users which weren't from the logged in server since the search by username requires @user@instance if it is a federated user and the comparison resulted in host being null.


jwr1 (approved):
Thanks. For a moment, I forgot the originInstance is actually the instance of the post/comment's author, so I sort of messed it up there.

PR #24, opened 2024-02-23
Lemmy user api

olorin99:
Adds lemmy api for users.
- get
- getByName
- updateProfile
- putBlock
- updateAvatar
- updateCover
- list posts from user
- list comments from user.

Since the lemmy api for user doesn't track page info I add a 'next_page' element to the json to imitate the list entries call.

### DIFF
--- a/lib/src/models/comment.dart
+++ b/lib/src/models/comment.dart
@@ -32,7 +32,7 @@ class CommentListModel with _$CommentListModel {
                   possibleChildren: json['comments'] as List<dynamic>,
                 ))
             .toList(),
-        nextPage: null,
+        nextPage: json['next_page'] as String?,
       );
 }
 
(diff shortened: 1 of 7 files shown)


jwr1 (approved):
Looks great; thanks!

lib/src/models/comment.dart
```diff
@@ -32,7 +32,7 @@ class CommentListModel with _$CommentListModel {
                   possibleChildren: json['comments'] as List<dynamic>,
                 ))
             .toList(),
-        nextPage: null,
+        nextPage: json['next_page'] as String?,
```
jwr1, line 35: ~@olorin99, I just checked the [lemmy comment api](https://lemmy.readme.io/reference/get_comment-list), and I don't think it actually has a `next_page` key.~

jwr1: Oh, never mind, I saw your comment about it. That's actually pretty smart; I was running into the same thing too.

PR #25, opened 2024-02-27
Implement search for lemmy.

olorin99:
Adds search for lemmy.


jwr1 (approved):
Thanks! Now there's only a few more things left I need to work on before we can release.

PR #33, opened 2024-04-16
Profile changes

olorin99:
Moves profile editing options onto separate screen.
Adds option to delete avatar and cover images.
Adds user settings.

### DIFF
--- a/lib/src/api/users.dart
+++ b/lib/src/api/users.dart
@@ -433,7 +433,44 @@ class APIUsers {
         httpErrorHandler(response, message: 'Failed to load site info');
 
         return UserSettings.fromLemmy((jsonDecode(response.body)['my_user']
-        ['local_user_view']['local_user']) as Map<String, Object?>);
+            ['local_user_view']['local_user']) as Map<String, Object?>);
+    }
+  }
+
+  Future<UserSettings> saveUserSettings(UserSettings settings) async {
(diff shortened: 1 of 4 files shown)


jwr1 (changes requested):
At first glance, I noticed two things. Wouold you be able to change all the `Switch`es in the profile settings to use [`SwitchListTile`](https://api.flutter.dev/flutter/material/SwitchListTile-class.html)s instead (to make it more consistent with our other settings. Here's example usage below:
```dart
SwitchListTile(
  title: const Text('Use Account Language Filter'),
  value: controller.useAccountLangFilter,
  onChanged: !isLemmy ? controller.updateUseAccountLangFilter : null,
),
```

Additionally, after changing a few of the switches, reloading the page seems to revert the switches back to how they were (in other words, it seems like toggling the switches isn't saving anything). It's also the same case with the image delete buttons (but maybe for those the API isn't working correctly :shrug:). Have the profile settings been working for your account correctly?

olorin99:
I had forgotten to add the notification settings to the save settings function. That should be fixed now.
Regarding deleting avatar/cover, those both worked for me so not sure what the issue was. I changed it so the delete buttons clear the corresponding image on the preview.

jwr1 (approved):
Thanks! It seems like everything is working now.

PR #45, opened 2024-05-22
Lemmy specific sorting

olorin99:
Adds lemmy sorting options for feed screen and magazines explore screen.


jwr1 (approved):
Just tried it out; looks great! Thanks.

PR #47, opened 2024-05-30
Post filters

olorin99:
Adds support for filtering posts in the main feed.
Main feed rejects posts where either the title or body match the filter regex.
Sub feeds (Users, Magazines, etc) are not filtered.


jwr1 (approved):
Thanks! I added a description to the settings so people know how to use it and made a few other minor changes, but otherwise, it looks great. I'll try to get the next update out sometime in the next few days.

PR #80, opened 2024-12-24
Fix bug when navigating to magazine feed when type action is set to t…

olorin99:
Found bug that when navigating to a magazine screen when the type action is set to tabs, that the magazine screen would display the main feed instead of the magazine feed.


jwr1 (approved):
Thank you!

PR #101, opened 2025-02-17
Add a script to build the AppImage with sharun and uruntime

### LINKED ISSUE #98: AppImage dosn't work well on most distros
### Describe the bug

The Interstellar AppImage used to work mostly well before there were any dependencies, but since the libmpv requirement was added (and likely others), the AppImage just doesn't work on most distros anymore without needing to install some dependencies. The script in the GH workflow already attempts to include shared dependencies that are required:

https://github.com/jwr1/interstellar/blob/5afc5421350e45ccd0c4c55019d3643b2852e097/.github/workflows/build_release.yml#L142-L148

I've looked into using [appimage-builder](https://github.com/AppImageCrafters/appimage-builder) (they have a [page for Flutter apps](https://appimage-builder.readthedocs.io/en/latest/examples/flutter.html)), but it seems like it might possibly be abandoned (last commit was over a year ago)? I also don't like that you have to use a command to generate a recipe for it, which means you have to rerun the command whenever a dependency might have changed.

There's also a new [go-appimagetool](https://github.com/probonopd/go-appimage/tree/master/src/appimagetool) which will detect and bundle all your application dependencies for you, but it does require a specific directory structure in order to work (not that big of a deal though). On Reddit, a samueru_sama discussed a [few issues with using go-appimagetool](https://www.reddit.com/r/linuxquestions/comments/1hk4wsz/comment/m3fmbwi) and recommended using [sharun](https://github.com/VHSgunzo/sharun) instead. Even one of the AppImage maintainers is also [considering using sharun](https://github.com/probonopd/go-appimage/issues/312). [Here's a list](https://github.com/pkgforge-dev/Anylinux-AppImages) of some "AnyLinux" AppImages that are built to run on any Linux systems and also use sharun (they could be used as examples).

Alternatively, instead of using any of these new toolings, we could stick with what we have and revise the script to not exclude dependencies from libflutter_linux_gtk.so, but to exclude from this [exclude list](https://github.com/AppImageCommunity/pkg2appimage/blob/master/excludelist) in the AppImageCommunity/pkg2appimage repo. Unfortunately, it seems like an absolute dependency (`/usr/lib/libmujs.so`) has ended up in Interstellar's executable, which could not be fixed by our current script.

Maybe @Samueru-sama would be willing to comment here with a recommendation?

### Version

0.7.2

### What platform(s) are you seeing the problem on?

Linux

jwr1:
Fixes #98.

Edit: Huge thanks to @Samueru-sama for providing examples (including a whole [demo script](https://github.com/user-attachments/files/18880201/Samueru-sama-interstellar-appimage-demo.tar.gz)) and troubleshooting everything with me! More context in #102.


linux/appimage/interstellar.desktop
```diff
@@ -1,6 +1,7 @@
 [Desktop Entry]
 Type=Application
 Name=Interstellar
-Exec=LD_LIBRARY_PATH=lib ./interstellar %u
+Exec=interstellar %u
 Icon=interstellar
 Categories=Network
+StartupWMClass=interstellar
```
Samueru-sama, line 7: You payed attention 👀

jwr1: Yup :) I wasn't even sure what it did at first; I had to look it up.

Samueru-sama:
Btw @jwr1 consider using archlinux as base which would let you use our [debloated packages](https://github.com/pkgforge-dev/llvm-libs-debloated) and likely make the final appimage less than 100 MiB.

In the demo I did not use uruntime and only used the smaller llvm package, I didn't include the iculess libxml which removes a 30 MiB libicudata lib from the AppImage.

jwr1:
For the moment, I'll stick with Ubuntu due to GH action's native support and the fact that my workflow is already setup for it.

I'll definitely keep it in mind for the future though. How does building work for Arch Linux with multiple architectures anyway? I thought they only supported x86_64.

Samueru-sama:
> For the moment, I'll stick with Ubuntu due to GH action's native support and the fact that my workflow is already setup for it. I'll definitely keep it in mind for the future though.

I think I can get someone to do a PR with the changes 👀

> How does building work for Arch Linux with multiple architectures anyway? I thought they only supported x86_64.

[We have arm docker images](https://github.com/pkgforge-dev/docker-archlinux) There is an archlinux arm fork, although often times the PKGBUILDs from archlinux itself also work without needing modifications beyond just changing x86_64 for aarch64 xd

[See example with arm](https://github.com/psadi/ghostty-appimage/blob/main/.github/workflows/ci.yaml)

> And, if I did decide to use Arch instead for the AppImage, would it be better to also compile the Linux build with Arch as well, or would it not matter?

Not really needed

> If I did use your debloated packages, I could also use the mini ffmpeg; then, I'd end up using all three of the debloated packages.

Cool

jwr1:
> I think I can get someone to do a PR with the changes 👀

I'd be fine with that. Maybe it could be added as a separate job. The only annoying thing I can think of is that I'd then have to maintain two sets of dependencies, one for the Linux build job and one for the AppImage job. And what happens if one of the Linux builds requires an older version of a library (due to being built on Ubuntu lts), and Arch Linux no longer supports it? I've had this happen before with Ubuntu using libmpv1 but Arch only had libmpv2, so the app wouldn't start up anymore.

jwr1:
Maybe it would just be better to compile the Linux build with Arch as well.

Samueru-sama:
I doubt that's going to be a problem, worse case scenario you can get the old lib from the archlinux archive, but switching build archlinux as well it is then 👀

PR #102, opened 2025-02-17
appimage builds

Samueru-sama:
WIP, haven't tested it.

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -78,7 +78,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libgtk-3-dev libx11-dev pkg-config cmake ninja-build libblkid-dev libmpv-dev mpv
           # AppImage only dependencies
-          sudo apt-get install -y xvfb x11-xserver-utils
+          sudo apt-get install -y xvfb x11-xserver-utils libgl1-mesa-dev libglu1-mesa-dev libglx-mesa0
 
       - name: Setup Flutter
         uses: subosito/flutter-action@v2
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -78,7 +78,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libgtk-3-dev libx11-dev pkg-config cmake ninja-build libblkid-dev libmpv-dev mpv
           # AppImage only dependencies
-          sudo apt-get install -y xvfb x11-xserver-utils libgl1-mesa-dev libglu1-mesa-dev libglx-mesa0
+          sudo apt-get install -y xvfb x11-xserver-utils libgl1-mesa-dev libglu1-mesa-dev libglx-mesa0 patchelf strace
 
       - name: Setup Flutter
         uses: subosito/flutter-action@v2
--- a/scripts/build-appimage.sh
+++ b/scripts/build-appimage.sh
@@ -27,8 +27,6 @@ wget "$LIB4BN" -O "$BUILD_DIR"/lib4bin
 chmod +x "$BUILD_DIR"/lib4bin
 xvfb-run -a -- "$BUILD_DIR"/lib4bin l -p -v -e -r -k -d "$BUILD_DIR"/AppDir "$EXE_BUNDLE_DIR"/interstellar
 
-find "$BUILD_DIR"/AppDir/shared -type f -exec strip -s -R .comment --strip-unneeded {} ';'
-
 # Prepare sharun
 ln -s ./bin/interstellar "$BUILD_DIR"/AppDir/AppRun
 "$BUILD_DIR"/AppDir/sharun -g
--- a/scripts/build-appimage.sh
+++ b/scripts/build-appimage.sh
@@ -12,7 +12,7 @@ LIB4BN="https://github.com/VHSgunzo/sharun/releases/latest/download/sharun-$ARCH
 URUNTIME="https://github.com/VHSgunzo/uruntime/releases/latest/download/uruntime-appimage-dwarfs-$ARCH"
 
 # Prepare AppDir
-mkdir -p "$BUILD_DIR"/AppDir/usr/share/applications "$BUILD_DIR"/AppDir/shared/lib
+mkdir -p "$BUILD_DIR"/AppDir/usr/share/applications
 
 cp -r "$EXE_BUNDLE_DIR"/data "$BUILD_DIR"/AppDir
 cp linux/appimage/interstellar.desktop "$BUILD_DIR"/AppDir
@@ -20,12 +20,12 @@ cp linux/appimage/interstellar.desktop "$BUILD_DIR"/AppDir/usr/share/application
 cp assets/icons/logo.png "$BUILD_DIR"/AppDir/interstellar.png
 cp assets/icons/logo.png "$BUILD_DIR"/AppDir/.DirIcon
 
-ln -s ./usr/share "$BUILD_DIR"/AppDir
-
 # ADD LIBRARIES
 wget "$LIB4BN" -O "$BUILD_DIR"/lib4bin
 chmod +x "$BUILD_DIR"/lib4bin
-xvfb-run -a -- "$BUILD_DIR"/lib4bin l -p -v -e -r -k -d "$BUILD_DIR"/AppDir "$EXE_BUNDLE_DIR"/interstellar
+xvfb-run -a -- "$BUILD_DIR"/lib4bin l -p -v -e -k \
+	-d "$BUILD_DIR"/AppDir \
+	"$EXE_BUNDLE_DIR"/interstellar
 
 # Prepare sharun
 ln -s ./bin/interstellar "$BUILD_DIR"/AppDir/AppRun
--- a/scripts/build-appimage.sh
+++ b/scripts/build-appimage.sh
@@ -23,7 +23,7 @@ cp assets/icons/logo.png "$BUILD_DIR"/AppDir/.DirIcon
 # ADD LIBRARIES
 wget "$LIB4BN" -O "$BUILD_DIR"/lib4bin
 chmod +x "$BUILD_DIR"/lib4bin
-xvfb-run -a -- "$BUILD_DIR"/lib4bin -d "$BUILD_DIR"/AppDir -p -v -e -s -k \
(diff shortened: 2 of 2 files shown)


Samueru-sama:
Does this application actually work where it is build? 👀

Because I'm seeing on strace that it is trying to find `libGLX.so.1`

image

And it seems such library is actually not present in ubuntu, because I even explicitly told lbi4bin to bundle all of `libGL` and it wasn't packaged. 

renaming `libGLX.so.0` to `libGLX.so.1` causes an undefined symbol error: 

image

Another issue might e that the app refuses to run in the CI, this is specially common with web apps that don't let you run them as root, hence lib4bin can't catch all the libs

Samueru-sama:
Just tried it in an ubuntu 24.04 container, same ubuntu version where this was built and same undefined symbol error. 🤔

This should work because the container has all the libraries that the application needs in case it they were found in the AppImage, the fact that it doesn't makes me wonder if the app actually works.

jwr1:
> Does this application actually work where it is build? 👀

Yes, the app definitely works. It's built with GH actions, and the builds are published on [GH releases](https://github.com/jwr1/interstellar/releases). As long as the correct dependencies are installed, then the [interstellar-linux-x86_64.tar.gz](https://github.com/jwr1/interstellar/releases/download/v0.7.2/interstellar-linux-x86_64.tar.gz) file with the app will open. As proof, you could install the Flathub version, which directly uses the .tar.gz file I mentioned and should work on any distro.

jwr1:
I'll trigger a manual build for the AppImage through GH actions just to be sure.

Samueru-sama:
alright it actually works if I launch the binary alone on ubuntu 24.04. 

Now the million dollar question is why this tries to find a non existing library when ran with sharun, and more importantly why it results in an undefined symbol error if I give it such lib.

Samueru-sama:
Good news I was able to fix the issue it seems. 

image

Bad news I have to write a lot of shell now xd

Samueru-sama:
there is a lot digest here, somehow the binary ignores rpath 💀

Samueru-sama:
This is fucked up, you actually ran into that issue already 

https://github.com/jwr1/interstellar/blob/main/linux/appimage/interstellar.desktop#L4

**That `LD_LIBRARY_PATH=lib` should not be needed at all**, the binary already has an rpath of `$ORIGIN/lib`, in all my years of linux this is the first time I see this.

jwr1:
Yeah, I'm no expert working with shared libraries or executables at a low level, but I had to add that line to get it to work initially. I believe other Flutter projects have done the same as well.

And, I just want to say thanks for looking into this. This all seems way more complicated than it should be, and there's no way I could have done it myself. I think the documentation on appimage.org needs to be updated at some point.

Samueru-sama:
Done 😁

image

https://github.com/Samueru-sama/interstellar-appimage-demo
https://github.com/Samueru-sama/interstellar-appimage-demo/releases/tag/continuous

I gave up dealing with ubuntu, because it still didn't work after solving the issue, set up that repo in less than 30 min and got all working. 

It is up to you if you want to merge those changes to here, I'm not very good with touching others people existing CI. 

>This all seems way more complicated than it should be, 

It really isn't complicated, the issue is that one doesn't expect that something as basic as rpath would not work here. 

This is like writing a shell script where grep doesn't work and you don't know it xd.

There are several different ways to make an application relocatable: 

* Build it statically, impossible here. ❌

* Ship the binary with the dynamic linker along with the libs it needs, the binay is launched by the dynamic linker and you add library dirs with the `--library-path` flag, this is what sharun does and turns out does not work here somehow ❌

* patch a relative interpreter and set `$LD_LIBRARY_PATH` to the bundled libraries, **thank god the relative interpreter patching actually works here** This is the solution I used ✅

* spin a namespace, which is a big problematic because some distros, most notably ubuntu disable user namespaces. ❌

Closing this since the entire CI needs major changes to get this to work.

jwr1:
Thank you so much for the help! I'll take a look at the demo when I get a chance.

It's unfortunate that it won't work on Ubuntu.

Samueru-sama:
> Thank you so much for the help! I'll take a look at the demo when I get a chance.
> 
> It's unfortunate that it won't work on Ubuntu.

[I was able to fix the issue after I made a bug report at sharun](https://github.com/Samueru-sama/interstellar-appimage-demo/commit/b6564f2ac0c9c5c1a51d88b20cb62ee860e29f91)

One of the very first things the "binary" does is load the relative `lib/libapp.so` library, which seems this is the actual application and not the binary, this would explain the issue with rpath being ignored. 

So yeah, you can just use the script on the repo as reference to make the AppImage. 

And tbh ubuntu is not a right distro to build packages anymore, it tends to split packages into a million pieces and somehow I still run into some missing packages or libraries. 

Or archlinux you just install mesa and you get all the opengl libs you need, [the version of mesa that ubuntu 24.04 ships is alsod lacking this very important optimization.](https://www.phoronix.com/news/Mesa-24.2-Modern-Interfaces)

And ubuntu 22.04 and older have a bug in mesa that prevent these appimages from using the proprietary nvidia driver.

I was able to convince the owner of the [ghostty appimage](https://github.com/psadi/ghostty-appimage) to switch building on archlinux to simplify the entire thing, not only it was a success, the author himself switched to using archlinux in his personal PC after 10 years of ubuntu 😁

jwr1:
I already use Arch on most of my systems btw. I was just hoping to get the build script working with Ubuntu since that's what GH actions uses.

jwr1:
I noticed in your demo script that you used appimagetool instead of runtime to bundle the AppImage. Is that what you would recommend using?

jwr1:
Also, what is the use in exporting these two variables?

```
export ARCH="$(uname -m)"
export APPIMAGE_EXTRACT_AND_RUN=1
```

Samueru-sama:
> I noticed in your demo script that you used appimagetool instead of runtime to bundle the AppImage. Is that what you would recommend using?

uruntime will make a smaller appimage indeed, I didnt bother because I wrote that quickly and that way I don't have to install llvm in the CI to add the update information to the runtime. 

Notice that I used this version of the llvm libs that is considerably smaller, this made the final appimage about the same size as the old appimage that didn't bundle everything 👀

So if you want to use the uruntime you can, same way I have some other debloated packages, like the iculess libxml which removes a 30 MiB libicudata lib dependency.

Samueru-sama:
> Also, what is the use in exporting these two variables?
> 
> ```
> export ARCH="$(uname -m)"
> export APPIMAGE_EXTRACT_AND_RUN=1
> ```

Those are for appimagetool.

jwr1:
Do you think your stripped-down llvm libs would work fine with an Ubuntu build?

Samueru-sama:
> Do you think your stripped-down llvm libs would work fine with an Ubuntu build?

Don't think so because the package is meant for archlinux.

jwr1:
Just a few more things:

- Trying to open a browser link from the app with my Ubuntu build fails with the following error. I was hoping you might have an idea of what the issue is. (It seems like the issue might not be present in your demo's arch build)
```
(one.jwr.interstellar:1701162): Gdk-WARNING **: 16:05:19.985: Server is missing xdg_foreign support
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: PlatformException(Launch Error, Failed to launch URL: Failed to execute child process “gio-launch-desktop” (No such file or directory), null, null)
#0      UrlLauncherLinux.launchUrl (package:url_launcher_linux/url_launcher_linux.dart:58)
<asynchronous suspension>
```
- Starting the AppImage with a uruntime build logs the following everytime the app starts, whereas when built with the appimagetool, it does not. Is this something I need to worry about?
```
Ignoring invalid max threads value 4294967295 > max (100000).
```
- Just a question here, how do you determine the optimal `-S` value when using uruntime? Is there any documentation that explains what it does anyway? When searching your pkgforge org, I noticed your projects used a range of values including `-S20`, `-S22`, `-S23`, `-S24`, and `-S25`.

Samueru-sama:
> * Trying to open a browser link from the app with my Ubuntu build fails with the following error. I was hoping you might have an idea of what the issue is. (It seems like the issue might not be present in your demo's arch build)

Can you link it? 

>Starting the AppImage with a uruntime build logs the following everytime the app starts, whereas when built with the appimagetool, it does not. Is this something I need to worry about?

[Known harmless issue with the newer version of fuse used by dwarfs](https://github.com/libfuse/libfuse/issues/1003)

>Just a question here, how do you determine the optimal -S value when using uruntime? Is there any documentation that explains what it does anyway? When searching your pkgforge org, I noticed your projects used a range of values including -S20, -S22, -S23, -S24, and -S25.

The value controls the block size, the higher the blocksize the smaller the appimage but it also increases the block size. 

I set the block size so that the launch time matches the launch time of the type2-runtime, if you you are not sure use -S22.

jwr1:
> Can you link it?

Yes, here it is: https://github.com/jwr1/interstellar/releases/download/v0.7.2/TEST-interstellar-linux-x86_64-UbuntuBuild.AppImage

Samueru-sama:
> > Can you link it?
> 
> Yes, here it is: https://github.com/jwr1/interstellar/releases/download/v0.7.2/TEST-interstellar-linux-x86_64-UbuntuBuild.AppImage

Looks like this does not use `xdg-open` to open links in default applications 😫

Cannot see anything in strace that would indicate why it doesn't work. going to ask for help.

Samueru-sama:
Alright here is the solution. 

You have to place the following script named `gio-launch-desktop` in top level `bin` dir of the AppImage: 

```
#!/bin/sh

if ! command -v gio-launch-desktop 1>/dev/null; then
	/usr/lib/gio-launch-desktop "$@"
fi
```

 The above script needs a lot of changes to handle the different locations distros might have the `gio-launch-desktop` binary. Ideally it should be turned to a wrapper to `xdg-open` which there likely is one already that could be used.

Samueru-sama:
Update, this is much simpler, replace the script for this instead:

```
#!/bin/sh

shift
xdg-open "$@"
```

jwr1:
Looks like that worked! It still logs `Server is missing xdg_foreign support` every time a browser link is opened, but the fact that it works is good enough for me; most people won't be looking at logs anyway.

Thanks again for all the help! I believe the AppImage is now truly ready for use on any distro (as opposed to the half-baked support we currently have).

PR #118, opened 2025-03-16
Update navigation handling so pages keep state when switching between them.

olorin99:
Should keep state when switching between pages on the navigation bar as well as between feed tabs.
Adds a callback to refresh widgets state when switching between accounts.

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -27,7 +27,9 @@ class App extends StatefulWidget {
 class _AppState extends State<App> {
   int _navIndex = 0;
   final PageController _pageController = PageController();
-  Key _refreshKey = UniqueKey();
+  Key _feedKey = UniqueKey();
+  Key _exploreKey = UniqueKey();
+  Key _accountKey = UniqueKey();
 
   void _changeNav(int newIndex) {
     setState(() {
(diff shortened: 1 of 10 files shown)


lib/src/app.dart
```diff
@@ -189,12 +197,24 @@ class _AppState extends State<App> {
...
+                        ],
+                      )
+                    ),
+                    // Expanded(
+                    //     child: const [
+                    //   FeedScreen(),
+                    //   ExploreScreen(),
+                    //   AccountScreen(),
+                    //   SettingsScreen(),
+                    // ][_navIndex]),
```
jwr1: Probably don't need to leave this part.

jwr1 (changes requested):
Thank you! Just a few points:

- Going to the settings and switching accounts seemed to cause a weird state where the screen would switch back to the feed view (without pressing anything) while the navigation would still have the settings highlighted. Ideally, it just wouldn't switch at all.
- Could you also have the state refresh on profile switch? There are a lot of default settings that could potentially change when the profile is switched, so a refresh would be needed in order to inherit those changes.

jwr1:
Maybe have the profile state refresh in `_rebuildProfile`.

lib/src/app.dart
olorin99: Lol oops.

PR #132, opened 2025-03-28
Fix being able to scroll between navigation screens.

olorin99:
Fixes bug where you could scroll between the navigation screens.


jwr1 (approved):
Thanks!

PR #133, opened 2025-03-28
Swipe actions

olorin99:
Adds customizable swipe gestures to content items. Currently available actions include
- upvote
- downvote
- boost
- bookmark
- reply
- moderatePin
- moderateMarkNSFW
- moderateDelete
- moderateBan


jwr1 (changes requested):
Looks great so far, just have a few points:

- Could you change the way it's stored in the profile to be four separate fields, instead of an array? I don't think we'd ever have any more or less than four options, so I think it's fine to do it that way. Changing it away from an array would also mean you wouldn't have to use those `var newActions = List<SwipeAction>.from`  lines in multiple places.
- Could you change the names of the settings to make it more obvious which gesture it's changing? Like use "Left" and "Right" in the names. Not exactly sure how to differentiate the ones that are on the same side though. Maybe "Short" and "Long"? And then also use the new name as the key in the profile as well.
- Could you make it so the autofocus in the markdown editor is an optional parameter? I certainly agree it should be auto focused in certain contexts, like while replying, but in others, like creating a post, there are other fields as well, and it shouldn't really be automatically focused.
- Also, I like the upvote and boost being on the left, but could you have a bookmark (short) and reply (long) on the right as the default?

jwr1 (changes requested):
I know this would be more work, but could you also add the swipe actions to the `ContentItemCompactPost` widget as well, as that's the kind of context where it would really be best in.

jwr1 (approved):
Thank you! It's looking great.

PR #143, opened 2025-04-01
Update haptics to use HapticFeedback class.

olorin99:
This felt good on my phone but may be different on others.
If we want more control I also saw https://pub.dev/packages/vibration which allows direct control of vibration variables like the intensity etc rather than just using presets.


jwr1 (approved):
Thank you! Still seems good on my device.

jwr1:
> If we want more control I also saw [pub.dev/packages/vibration](https://pub.dev/packages/vibration) which allows direct control of vibration variables like the intensity etc rather than just using presets.

That seems like quite an interesting package. I could see us using it in the future.

jwr1:
I almost forgot the android vibrate permission, but I have added it.

Issue #130, opened 2025-03-28
Add Local View


wjs018:
### Describe the feature

I would love to have a local view, in which you see all the posts from your home instance's communities. I tend to use lemmy-ui's local view with instances that are focused on a specific topic instead of subscribing to all the communities on that instance.

For example, let's say I am a big fan of Star Trek. I would create an account on the startrek.website lemmy instance, and then just browse the `local` posts instead of subscribing to each community on the instance. This also makes it easier to discover new communities that are created on the instance.

jwr1:
Complete in db9e1743405ad46eac134c4ed16ab1f301347b87.

I'll wait till https://github.com/MbinOrg/mbin/issues/1528 is resolved before closing, though, to ensure the API was implemented accurately.

jwr1:
It would probably just be best to hide the Local view on Mbin for now (but add a comment in the code), so it can be released in this next update.

jwr1:
Closed in #145.

jwr1:
The Mbin side is now complete for this.

Interstellar: ef4d882040b211c73865fcda95d3ef899370673a
Mbin: https://github.com/MbinOrg/mbin/pull/1618

PR #145, opened 2025-04-04
Hides the local filter on mbin

olorin99:
Disables local filter on mbin with TODO comment reminding to remove once added to mbin as suggested [here](https://github.com/jwr1/interstellar/issues/130#issuecomment-2764819741)


jwr1 (approved):
Thank you!

PR #146, opened 2025-04-06
Piefed - enable handling user about for piefed

JollyDevelopment:
This adjusts the `api.users.updateProfile()` for Piefed. It matches the below PR on the Piefed project that adds the `api/alpha/user/save_user_settings` route.

Piefed side PR: https://codeberg.org/rimu/pyfedi/pulls/545

So far i've tested and it works on the Linux debug app. Going to build an android app as well and make sure it works there.

Both PRs are draft at the moment as I make sure it actually works. :-)

### DIFF
--- a/lib/src/api/users.dart
+++ b/lib/src/api/users.dart
@@ -222,7 +222,14 @@ class APIUsers {
         return null;
 
       case ServerSoftware.piefed:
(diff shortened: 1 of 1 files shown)


jwr1 (approved):
Thanks!

PR #147, opened 2025-04-07
Move overflow menu into bottom modal sheet.

### LINKED ISSUE #140: Notification controls for PieFed comments take up too much space
### Describe the bug

Notification controls for PieFed comments take up too much space by adding a whole extra row to the UI.

### Steps to Reproduce

_No response_

### Version

0.9.0

Linux, Android

PieFed

### Additional context

See this comment for context: https://kbin.earth/m/piefed_meta@piefed.social/t/1099836/Interstellar-s-latest-update-supports-PieFed-s-alpha-API/comment/5828401#entry-comment-5828401

olorin99:
Move overflow menu into bottom modal sheet.
Also added notification controls and voting controls with separate upvotes/downvote counts to top of new modal menu.
Closes #140.


jwr1 (changes requested):
Thank you, this looks great! Just two things I've noticed:

- Clicking on the various actions (e.g., upvoting, bookmarking, etc.) does not update the UI in the popup modal (upvoting does not change the number or arrow color). There are two ways to solve this: either update the state, or close the popup when an action is performed. Closing the popup once an action is performed would match it up with the way it worked previously, so that might be the best way to go.
- Again, I know this means more work, but could you update ContentItemCompactPost, as that's where this sort of popup would really be helpful (as there are no other controls).

olorin99:
Since they share a lot of functionality I moved the compact content item into a function in the full content item and which widget structure is returned in build is chosen based on isCompact. This should be easier to keep action callbacks consistent between them with just the ui changing.

jwr1 (changes requested):
Combining the two content item's into the same class is a good call. I just have three last things:

- Now that you can long tap over the post to pop up the menu, can you make it so right clicking does the same (long tap and right click are paired in several other places as well)?
- Opening the full post view still has the ink well there, even though tapping on it doesn't do anything. So, probably make the long tap (and right click) only work if the post is in the feed (not full screen).
- The ban user dialog no longer works, due to `Navigator.pop` being used immediately after and closing the dialog. Moving `pop` to before running the ban action should fix it.

jwr1 (approved):
LGTM!

PR #148, opened 2025-04-07
bump dwarfs block size and dedup

Samueru-sama:
This recently got a lot of performance improvements thanks to the dwarfs developer finding out what we were doing 👀

image

Also added missing `.DirIcon` 😅


jwr1 (approved):
Oh nice, thanks!

PR #150, opened 2025-04-10
Add time options for commented sort.

olorin99:
Adds time options for commented sort.
Adds missing localization strings for sort times.


jwr1 (approved):
Thank you! This is great. It was top sort and commented sort that really needed those time frames in order to be useful.

PR #152, opened 2025-04-11
Update comment collapsing.

olorin99:
Remove collapse comment button.
Tap collapses the comment if onClick callback is not provided otherwise uses onClick (on explore/users screens).
Long press and right click open the content menu similar to posts on the feed screen.
Collapsing comment collapses comment text as well as children. 
Add colours to comment thread bands.


jwr1 (approved):
Thank you, it looks great!

PR #153, opened 2025-04-13
ci: Separate Linux Builds for AppImages

### LINKED ISSUE #129: Proposed CI workflow changes for Interstellar (Specific to AppImage Builds)
### Describe the feature

Continuing from #101,

Based on my CI analysis, the existing workflow is as follows:

1. Create draft release 
2. Create build (matrix for windows, Linux (x86, arm), android)
3. Build appimage (calling build script) / build with some customisation based on the runner platform
4. Publish release with built artifacts

**Proposed Changes.**

I suggest we decouple the AppImage build process from the current workflow and implement a separate downstream workflow that triggers after the completion of `build_release` . In this new workflow, we can utilise the [pkgforge Arch image](https://github.com/pkgforge-dev/docker-archlinux) to build the AppImage with debloated packages + other customisations and publish the appimage artifacts to the existing release

The current workflow cannot accommodate this change because the existing matrix strategy uses a Windows runner, which would conflict with the introduction of the Arch image.

Please share your thoughts on this proposal. If you agree, I’d be happy to contribute the changes.

### Additional context

_No response_

psadi:
This commit introduces the following changes:

- Utilized the pkgforge Arch Linux image for x86_64 and aarch64 builds
- Manually resolved build dependencies & employ pkgforge debloated packages (llvm, libxml, and ffmpeg) to minimize overall build size
- Updated uruntime build parameters for improved performance
- Added UPINFO for automatic AppImage updates

Closes #129

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -30,7 +30,7 @@ jobs:
           tag_name: v${{ inputs.version }}
           draft: true
 
-  build-android-windows:
+  create-build:
     needs: draft-release
     environment: production
     name: Create ${{ matrix.target }} build
@@ -38,6 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        target: [android, linux-x86_64, linux-aarch64, windows-x86_64]
         include:
           - target: android
             target_os: android
@@ -45,12 +46,25 @@ jobs:
             build_flags: --split-per-abi
             build_path: build/app/outputs/flutter-apk
             runner: ubuntu-latest
+          - target: linux-x86_64
+            target_os: linux
+            target_arch: x86_64
+            build_target: linux
+            build_path: build/linux/x64/release/bundle
+            runner: ubuntu-24.04
+          - target: linux-aarch64
+            target_os: linux
+            target_arch: aarch64
+            build_target: linux
+            build_path: build/linux/arm64/release/bundle
+            runner: ubuntu-24.04-arm
           - target: windows-x86_64
             target_os: windows
             target_arch: x86_64
             build_target: windows
             build_path: build\windows\x64\runner\Release
             runner: windows-latest
+    container: ${{ matrix.target_os == 'linux' && 'ghcr.io/pkgforge-dev/archlinux:latest' || null }}
     steps:
       - name: Install android dependencies
         if: matrix.target_os == 'android'
@@ -59,6 +73,14 @@ jobs:
           java-version: "21"
           distribution: temurin
 
+      - name: Install linux dependencies
+        if: matrix.target_os == 'linux'
+        run: |
+          pacman -Syuq --needed --noconfirm --noprogressbar \
+            ninja gtk3 xz gcc mpv wget jq git which base-devel \
+            file zsync patchelf binutils strace mesa llvm \
+            xorg-server-xvfb cmake clang unzip
+
       - name: Install windows dependencies
         if: matrix.target_os == 'windows'
         uses: ilammy/setup-nasm@v1
@@ -66,13 +88,17 @@ jobs:
       - name: Setup Flutter
         uses: subosito/flutter-action@v2
         with:
-          channel: stable
+          # Use master until Flutter action is fixed for arm (https://github.com/subosito/flutter-action/issues/345)
+          channel: ${{ matrix.target == 'linux-aarch64' && 'master' || 'stable' }}
           flutter-version: 3.29.0
 
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          persist-credentials: false
+
+      - name: git-config add safe directory
+        if: ${{ matrix.target_os == 'linux' }}
+        run: |
+          git config --global --add safe.directory $FLUTTER_ROOT
 
       - name: Set version
         run: |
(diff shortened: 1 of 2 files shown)


jwr1:
Thank you for taking the time to do this!

Correct me if I'm wrong, but the way you've done it, it sort of seems like the Linux job would still work fine if it were still combined back into the original matrix, the big difference being that the container field is set in the job.

jwr1:
Couldn't we have the `container` field set only while the matrix OS is Linux?

psadi:
> Couldn't we have the `container` field set only while the matrix OS is Linux?

The container option is a higher-level input (similar to environment and strategy) and to my knowledge, there isn't a built-in way to handle this conditionally. We can only select the image based on the matrix, which is why we separate Linux builds into their own job as the container field is not necessary for Android or Windows.

jwr1:
I just created a test workflow, and the following line seems to work as expected:

```yaml
container: ${{ matrix.target_os == 'linux' && 'ghcr.io/pkgforge-dev/archlinux:latest' || null }}
```

psadi:
> I just created a test workflow, and the following line seems to work as expected:
> 
> ```yaml
> container: ${{ matrix.target_os == 'linux' && 'ghcr.io/pkgforge-dev/archlinux:latest' || null }}
> ```
> 
> When the job was one of the two linux targets (x86 or arm), it used the Arch container, otherwise, it did not.

Could you test by including windows and android runners as well and share the outcome here ?

jwr1:
> > I just created a test workflow, and the following line seems to work as expected:
> > ```yaml
> > container: ${{ matrix.target_os == 'linux' && 'ghcr.io/pkgforge-dev/archlinux:latest' || null }}
> > ```
> > When the job was one of the two linux targets (x86 or arm), it used the Arch container, otherwise, it did not.
> 
> Could you test by including windows and android runners as well and share the outcome (specifically looking if the steps ran within the defined runner)

I actually did that as well. Here's the full workflow I tested with, and all of the jobs succeded:

```yaml
name: Build release

permissions:
  contents: write

on:
  workflow_dispatch:

jobs:
  build:
    name: Create ${{ matrix.target }} build
    runs-on: ${{ matrix.runner }}
    strategy:
      fail-fast: false
      matrix:
        include:
          - target: android
            target_os: android
            runner: ubuntu-latest
          - target: linux-x86_64
            target_os: linux
            target_arch: x86_64
            runner: ubuntu-24.04
          - target: linux-aarch64
            target_os: linux
            target_arch: aarch64
            runner: ubuntu-24.04-arm
          - target: windows-x86_64
            target_os: windows
            target_arch: x86_64
            runner: windows-latest
    container: ${{ matrix.target_os == 'linux' && 'ghcr.io/pkgforge-dev/archlinux:latest' || null }}
    steps:
      - name: "Hello ${{ matrix.target }}!" 
        run: echo "Hello ${{ matrix.target }}!"
```

After the first test, I added a `pacman --version` step in there, and the two linux jobs displayed the version, and the other two failed, as expected.

psadi:
I've implemented the changes to the single job as you suggested. I've tested it and the logic is functioning correctly. This should make it easier for you to maintain in the future. Please take a look and share your thoughts.

psadi:
> Also, how does UPINFO work?

`UPINFO` is a metadata that can be included in an AppImage. It contains essential information about the application, such as its version, the GH repo info for updates & other relevant details. The `zsyncmake` tool here generates an additional `.AppImage.zsync` file that references the `UPINFO` file along with the AppImage format. 

The main benefits here are,

1. **Automatic Updates:** When an AppImage is executed, it can check for updates by looking for the UPINFO file. If the file is present, the application can use the information within it to determine if a newer version is available. Package managers that distribute AppImages, such as [AM](https://github.com/ivan-hc/AM), [Soar](https://github.com/pkgforge/soar), and [AppImageUpdate](https://github.com/AppImageCommunity/AppImageUpdate), rely on this information to automate the update process.
2. **User Notification:** If an update is found, the application (if it has the necessary capabilities) can notify the user and provide options to download and install the new version.
3. **Cross-Distribution Compatibility:** As you might already know that one of the main advantages of AppImage is its design to work across various Linux distributions. The `.AppImage.zsync` file helps maintain this compatibility by allowing applications to manage their updates independently of the underlying system.

.github/workflows/build_release.yml
```diff
@@ -92,6 +95,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: git-config add safe directory
+        if: ${{ matrix.target_os == 'linux' }}
+        run: |
+          git config --global --add safe.directory $FLUTTER_ROOT
```
jwr1, line 101: What does this do?

scripts/build-appimage.sh
```diff
@@ -40,12 +41,17 @@ ln "$BUILD_DIR"/AppDir/sharun "$BUILD_DIR"/AppDir/AppRun
...
-	--compression zstd:level=22 -S26 -B32 \
-	--header "$BUILD_DIR"/uruntime \
-	-i "$BUILD_DIR"/AppDir -o dist/interstellar-linux-$ARCH.AppImage
+    --set-owner 0 --set-group 0 \
+    --no-history --no-create-timestamp \
+    --compression zstd:level=22 -S26 -B8 \
+    --header "$BUILD_DIR"/uruntime \
+    -i "$BUILD_DIR"/AppDir -o dist/interstellar-"$VERSION"-"$ARCH".AppImage
+
+zsyncmake dist/*.AppImage -u dist/*.AppImage -o "dist/interstellar-${VERSION}-${ARCH}.AppImage.zsync"
```
jwr1: Do the AppImage filenames have to include the app version in them? Would it be possible to remove the version so it matches the format of all the other release files?

jwr1 (changes requested):
Thanks for the update! I just have two quick questions.

.github/workflows/build_release.yml
psadi: Since we run in arch container, Set Version tasks fails with the following error

```
Run flutter pub global activate cider
  flutter pub global activate cider
  cider version 1.2.3+1
  shell: sh -e {0}
  env:
    FLUTTER_ROOT: /__t/flutter/stable-3.29.0-x64
    PUB_CACHE: /github/home/.pub-cache
fatal: detected dubious ownership in repository at '/__t/flutter/stable-3.29.0-x64'
To add an exception for this directory, call:

	git config --global --add safe.directory /__t/flutter/stable-3.29.0-x64
Error: Process completed with exit code 128.
```

Ref (build) :https://github.com/psadi/interstellar/actions/runs/14450603540/job/40522096255#step:8:1

To handle this, the step was added to add unsafe directories only for Linux OS

scripts/build-appimage.sh
psadi: It is not mandatory to have the version included in the file name at-least looking at the update spec [here](https://github.com/AppImage/AppImageSpec/blob/master/draft.md#update-information). However i feel its generally a good idea to include the version for clarity and visibility (more like generic steps we follow for all AppImage builds). I leave this to your decision here.

jwr1: Yeah, I'd prefer that the filenames do not have the version in them then, like how it was previously (e.g., `interstellar-linux-x86_64.AppImage`).

.github/workflows/build_release.yml
jwr1: Ok, thanks for the info.

scripts/build-appimage.sh
psadi: [222ae93](https://github.com/jwr1/interstellar/commit/222ae93058e0dfdfe03c00606e3818f36299cf85) should accommodate the changes requested.

jwr1:
I just ran a test release build, and the AppImages ended up having about a 40MB size reduction :tada:

PR #161, opened 2025-04-22
Translate post option.

olorin99:
Add an option in the content menu to translate a selected post.
Also add an option in behaviour settings to auto translate posts in feed.
Translations are shown under the original body with the source and target languages.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -181,7 +181,7 @@
   "settings_askBeforeDeleting": "Ask before deleting",
   "settings_autoPlayVideos": "Auto play videos",
   "settings_hapticFeedback": "Haptic feedback",
-  "settings_autoTranslate": "Auto translate posts",
+  "settings_autoTranslate": "Auto translate posts/comments",
   "settings_themeMode": "Theme mode",
   "settings_themeMode_system": "System",
   "settings_themeMode_light": "Light",
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -181,6 +181,7 @@
   "settings_askBeforeDeleting": "Ask before deleting",
   "settings_autoPlayVideos": "Auto play videos",
   "settings_hapticFeedback": "Haptic feedback",
+  "settings_autoTranslate": "Auto translate posts",
   "settings_themeMode": "Theme mode",
   "settings_themeMode_system": "System",
   "settings_themeMode_light": "Light",
--- a/lib/src/widgets/content_item/content_item.dart
+++ b/lib/src/widgets/content_item/content_item.dart
@@ -32,7 +32,9 @@ class ContentItem extends StatefulWidget {
   final ImageModel? image;
   final Uri? link;
   final String? body;
+  final Translation? translation;
   final String? lang;
+  final Future<void> Function(String lang)? onTranslate;
   final DateTime? createdAt;
   final DateTime? editedAt;
 
@@ -106,7 +108,9 @@ class ContentItem extends StatefulWidget {
     this.image,
     this.link,
     this.body,
+    this.translation,
     this.lang,
+    this.onTranslate,
     this.createdAt,
     this.editedAt,
     this.isPreview = false,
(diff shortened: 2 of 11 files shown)


lib/l10n/app_en.arb
```diff
@@ -181,6 +181,7 @@
   "settings_askBeforeDeleting": "Ask before deleting",
   "settings_autoPlayVideos": "Auto play videos",
   "settings_hapticFeedback": "Haptic feedback",
+  "settings_autoTranslate": "Auto translate posts",
```
jwr1:
```suggestion
  "settings_autoTranslate": "Auto translate posts/comments",
```

jwr1 (changes requested):
This looks great! It's a pretty cool idea to have post/comment translating built right into the app.

A few things to note:

- I like the way auto translate works. I just think the setting toggle label should make it clear auto translate works for both posts and comments.
- Manual translate (through the overflow menu), opens up a dialog that has the source code of the original text on the top with the source code of the translated text on the bottom. It seems like making it work more like auto translate does would make it more useful, like instead of the dialog with the source code displaying, maybe close the overflow menu altogether so the post is visible again, and then display the rendered translation below the post like it does for auto translate. Presumable, if you run it multiple times (or after it was auto translated), it would just replace the current translation with the new one.

jwr1:
Will this work if a post/comment does not provide a language? I just realized that only while using Mbin (not Lemmy or PieFed) will posts/comments have a non-null `lang`.

lib/src/widgets/content_item/content_item.dart
```diff
@@ -164,12 +168,93 @@ class ContentItem extends StatefulWidget {
...
+            ],
+          )
+        : widget.isPreview
+        ? Text(
+            widget.body!,
+            maxLines: 4,
+            overflow: TextOverflow.ellipsis,
+          )
+        : Markdown(widget.body!, widget.originInstance);
+  }
```
jwr1: No code change here, just formatted it and added two comments to make it a bit more readable.

```suggestion
  Widget contentBody(BuildContext context) {
    return _translation != null
        // A translation is available
        ? widget.isPreview
            ? Column(
                crossAxisAlignment: CrossAxisAlignment.start,
                mainAxisSize: MainAxisSize.min,
                children: [
                  Text(
                    widget.body!,
                    maxLines: 4,
                    overflow: TextOverflow.ellipsis,
                  ),
                  Divider(),
                  Row(
                    mainAxisAlignment: MainAxisAlignment.spaceEvenly,
                    children: [
                      Text(getLanguageName(context, widget.lang!)),
                      Icon(Symbols.arrow_right),
                      Text(_translation!.targetLanguage.name),
                    ],
                  ),
                  Divider(),
                  Text(
                    _translation!.translations.text,
                    maxLines: 4,
                    overflow: TextOverflow.ellipsis,
                  ),
                ],
              )
            : Column(
                crossAxisAlignment: CrossAxisAlignment.start,
                mainAxisSize: MainAxisSize.min,
                children: [
                  Markdown(widget.body!, widget.originInstance),
                  Divider(),
                  Row(
                    mainAxisAlignment: MainAxisAlignment.spaceEvenly,
                    children: [
                      Text(getLanguageName(context, widget.lang!)),
                      Icon(Symbols.arrow_right),
                      Text(_translation!.targetLanguage.name),
                    ],
                  ),
                  Divider(),
                  Markdown(
                      _translation!.translations.text, widget.originInstance),
                ],
              )
        // No translation is available
        : widget.isPreview
            ? Text(
                widget.body!,
                maxLines: 4,
                overflow: TextOverflow.ellipsis,
              )
            : Markdown(widget.body!, widget.originInstance);
  }
```

olorin99:
Moved the translation itself out of content_menu into a callback from which the results are passed down into content_item so translations are always shown in the post/comment body and looks much better.
Yeah the auto translation requires the post/comment language to be correct. For example I've noticed most (if not all) posts from misskey are marked as english even though they're japanese. The translation api can auto detect the source language so manually triggering the translation will work.

PR #162, opened 2025-04-23
Add support for marking posts as read

### LINKED ISSUE #105: Mark post as read support for Lemmy
### Describe the feature

(Not possible on Mbin though)

### Additional context

_No response_

olorin99:
For lemmy just uses lemmy api to mark as read.
For all other account types read status is stored in a sembast store.
Posts are marked on read when opened or through swiping or the menu.
Also added feed actions to quickly show/hide read posts in the feed. 
Closes #105


lib/src/api/threads.dart
```diff
@@ -580,4 +580,24 @@ class APIThreads {
...
+      case ServerSoftware.mbin:
+        throw UnsupportedError('Mbin doesnt support marking posts as read');
+      case ServerSoftware.lemmy:
+        const path = '/post/mark_as_read';
+
+        final response = await client.post(
+          path,
+          body: {
+            'post_id': postId,
+            'post_ids': [postId],
```
jwr1, line 594: I believe you only need this line if you are marking multiple posts as read. Since we are only marking one and already use the `post_id` field, then `post_ids` is unnecessary. `post_ids` is useful if you are marking a lot of posts as read in quick succession, like if you have them marked as read as you scroll past them. Some Lemmy clients actually do have this option (to mark as read once scrolled past), and I could see it being a good option to add to Interstellar sometime in the future.

```suggestion
```

lib/src/screens/feed/feed_screen.dart
```diff
@@ -963,31 +1003,38 @@ class _FeedScreenBodyState extends State<FeedScreenBody>
...
-                        }),
-                        onTap: onPostTap,
-                        filterListWarnings: _filterListWarnings[(item.type, item.id)],
-                        userCanModerate: widget.userCanModerate,
-                        isTopLevel: true,
-                        isCompact: true,
+                      DecoratedBox(
+                        decoration: BoxDecoration(
+                          color: item.read != null && item.read!
+                              ? Theme.of(context).cardColor.lighten(10)
```
jwr1: It seems like it would make more sense to darken the color instead of lighten, in order to bring less attention to the posts marked as read. I tested for a few minutes and `darken(3)` seemed like it could work (need to check both Light and Dark mode). For reference, when a post is marked as read on Lemmy the title text color is darkened/greyed out.

lib/src/screens/feed/feed_screen.dart
```diff
@@ -997,6 +1044,9 @@ class _FeedScreenBodyState extends State<FeedScreenBody>
                   );
                 } else {
                   return Card(
+                    color: item.read != null && item.read!
+                        ? Theme.of(context).cardColor.lighten(10)
```
jwr1: Same as above about darkening.

lib/src/models/post.dart
```diff
@@ -76,6 +76,7 @@ class PostModel with _$PostModel {
     required bool? canAuthUserModerate,
     required NotificationControlStatus? notificationControlStatus,
     required List<String>? bookmarks,
+    required bool? read,
```
jwr1: Would it make sense to make this non-nullable, so you don't have to check `read != null` multiple times? Mbin and PieFed already use a non-null default (`false`), and Lemmy could fallback to `false` if `null` is provided. I don't know for sure, but it's possible Lemmy's `read` field is only `null` when you are a guest.

lib/src/controller/controller.dart
jwr1, line 1: With you adding the new db store here, since you hold a reference to a user account, then you'll need to add clean up code to the delete account method that removes all the old post reads.

lib/src/controller/controller.dart
```diff
@@ -575,4 +576,20 @@ class AppController with ChangeNotifier {
...
+    if (await isRead(postId)) return;
+    await _readStore.add(db, {'postId': postId, 'read': read, 'account': _selectedAccount});
+  }
+
+  Future<bool> isRead(int postId) async {
+    final records = (await _readStore.find(db, finder: Finder(
+      filter: Filter.equals('postId', postId)
+    )));
+
+    return records.where((record) => record.value['account'] == _selectedAccount).firstOrNull != null;
```
jwr1: Is it possible to merge/cleanup these two checks together, like change the db query to not just filter by `postId`, but to also filter by `account` at the same time?

lib/src/controller/profile.dart
jwr1, line 1: The new feed action (hide read posts) needs to be added to the profile and a new settings entry needs to be added as well (for selecting the action location), this makes it so the action is not hard coded to `fab` like it is now.

Additionally, since the new "Hide read posts" setting is directly related to this new feed action, maybe you could move it to the feed defaults settings area instead of behavior.

To outline, instead of one new `bool hideReadPosts` item here, it should be `bool feedDefaultHideReadPosts` and `ActionLocation feedActionHideReadPosts`, and then update the corresponding settings.

lib/src/widgets/content_item/content_menu.dart
```diff
@@ -168,6 +168,14 @@ void showContentMenu(
                     ),
                     onTap: () => showBookmarksMenu(context, widget),
                   ),
+                if (widget.onMarkAsRead != null)
+                  ListTile(
+                    title: Text(l(context).action_markRead),
```
jwr1: Could this text change based on whether the post is `read` or not? Like "Mark as read" if the post is unread, and "Mark as unread" if the post is read.

jwr1 (changes requested):
Thanks, looks nice! I did end up having a lot of comments this time though.

lib/src/api/threads.dart
olorin99: I thought so too and the lemmy API docs only have the `read` parameter as required but it throws
`ClientException: Request failed with status 400: Bad Request: {"error":"unknown","message":"Json deserialize error: missing field `post_ids` at line 1 column 32"}` if I remove it.

jwr1: Ah Ok, thanks for checking. One more question, would it work the other way around (removing `post_id` and keeping `post_ids`)?

olorin99: Yep works that way around. I'll leave `post_id` removed.

jwr1 (approved):

PR #167, opened 2025-04-30
Add options to hide the vote and or comment buttons on content items

### LINKED ISSUE #134: Add option to hide action buttons
### Describe the feature

As a follow up to https://github.com/jwr1/interstellar/issues/108, the following would be great:

- [x] Remove the collapse button and allow collapsing comments through other means. This could potentially be by long tapping on the comment (like other Lemmy mobile apps) or maybe tapping on the top part of the comment (Reddit style). #152
- [ ] Add the option to hide action buttons from posts and comments now that gestures have been implemented.

### Additional context

_No response_

olorin99:
Closes #134


jwr1 (changes requested):
Thanks! Just a few points:

- I can't really imagine a case where having the option to control both the vote widgets and the comment widgets separately would be useful, so I'd say the option to hide them should be combined. Maybe something like "Hide action buttons" (`hideActionButtons` in the profile) or something similar.
- I believe some sort of description/subtitle would also be helpful for this setting. Maybe something like:

  > Hides the action buttons (vote, boost, reply, etc.) below each post and comment. The controls are still accessible in the overflow menu. It is recommended to enable swipe gestures in unison with this setting.
- It looks like the icon for the new setting toggle needs to be updated, as it's just the same as the one above it.
- Can you also add the notification controls to hide when this settings is enabled?
- Can you add a reply button to the overflow menu? I'm just thinking if somebody did enable this setting but did not enable swipe gestures, then there'd be no way to reply to anything.

olorin99:
Moved the inkwell that opens the overflow menu into content_item since onReply sets the content item state. The reply action won't be available from a collapsed comments overflow menu but I figure users shouldn't be able to reply to hidden content anyway so its alright.

jwr1 (approved):

PR #169, opened 2025-05-03
Fix search input not losing focus when tapping elsewhere on screen.

olorin99:
Whenever I searched for something, opened a result then navigated back to the explore screen the text input would retake focus and reopen the keyboard. This should fix that.


jwr1 (approved):
I was having the same problem on Android. I had looked into it a while ago and was honestly unsure how to fix it. Thanks!

olorin99:
Praises be to that random user on stackoverflow who posted the exact fix.

PR #170, opened 2025-05-04
Notifications Support for Piefed

JollyDevelopment:
This adds support for notifications for Piefed.

In coordination with this PR on Piefed: https://codeberg.org/rimu/pyfedi/pulls/615 (now merged), and https://codeberg.org/rimu/pyfedi/pulls/619.

It's mostly based on the way the `.fromMbin` functions are setup, except the NotificationItem model which does a switch on the notificationtype more like the lemmy way.

I tested it on my personal piefed instance and its working so far.

Works for the Linux client and the android app in my testing.

### DIFF
--- a/lib/src/api/notifications.dart
+++ b/lib/src/api/notifications.dart
@@ -62,7 +62,13 @@ class APINotifications {
             nextPage: lemmyCalcNextIntPage(result.items, page));
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path =
+            '/user/notifications/${filter == NotificationsFilter.new_ ? 'new' : (filter?.name ?? 'all')}';
+        final query = {'p': page};
+
+        final response = await client.get(path, queryParams: query);
+
+        return NotificationListModel.fromPiefed(response.bodyJson);
     }
   }
 
@@ -158,7 +164,12 @@ class APINotifications {
         };
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path =
+            '/user/notifications/$notificationId/${readState ? 'read' : 'unread'}';
+
+        final response = await client.put(path);
+
+        return NotificationModel.fromPiefed(response.bodyJson);
     }
   }
 
--- a/lib/src/api/notifications.dart
+++ b/lib/src/api/notifications.dart
@@ -63,9 +63,10 @@ class APINotifications {
 
       case ServerSoftware.piefed:
         final path = '/user/notifications';
-        final status = filter == NotificationsFilter.new_
-            ? 'new'
-            : (filter?.name ?? 'all');
+        String status = 'All'; // default to All
+        if (filter == NotificationsFilter.new_) status = 'New';
+        if (filter == NotificationsFilter.read) status = 'Read';
+
         final query = {'page': page, 'status_request': status};
 
         final response = await client.get(path, queryParams: query);
@@ -174,13 +175,10 @@ class APINotifications {
 
       case ServerSoftware.piefed:
         final path = '/user/notification_state';
-        final notifId = notificationId;
-        final query = {
-          'notif_id': notifId.toString(),
-          'read_state': readState.toString()
-        };
 
-        final response = await client.put(path, queryParams: query);
+        final body = {'notif_id': notificationId, 'read_state': readState};
+
+        final response = await client.put(path, body: body);
 
         return NotificationModel.fromPiefed(response.bodyJson);
     }
--- a/lib/src/api/notifications.dart
+++ b/lib/src/api/notifications.dart
@@ -63,9 +63,11 @@ class APINotifications {
 
       case ServerSoftware.piefed:
         final path = '/user/notifications';
-        String status = 'All'; // default to All
-        if (filter == NotificationsFilter.new_) status = 'New';
-        if (filter == NotificationsFilter.read) status = 'Read';
+        final status = switch (filter) {
+          NotificationsFilter.new_ => 'New',
+          NotificationsFilter.read => 'Read',
+          NotificationsFilter.all || null => 'All',
+        };
 
         final query = {'page': page, 'status_request': status};
 
--- a/lib/src/api/notifications.dart
+++ b/lib/src/api/notifications.dart
@@ -93,7 +93,11 @@ class APINotifications {
             (response.bodyJson['private_messages'] as int);
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        const path = '/user/notifications_count';
+
+        final response = await client.get(path);
+
+        return response.bodyJson['count'] as int;
     }
   }
 
@@ -114,7 +118,11 @@ class APINotifications {
         return;
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        const path = '/user/mark_all_notifications_read';
+
+        final response = await client.put(path);
+
+        return;
     }
   }
 
--- a/lib/src/api/notifications.dart
+++ b/lib/src/api/notifications.dart
@@ -62,9 +62,11 @@ class APINotifications {
             nextPage: lemmyCalcNextIntPage(result.items, page));
 
       case ServerSoftware.piefed:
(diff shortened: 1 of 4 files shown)


lib/src/screens/account/notification/notification_screen.dart
```diff
@@ -199,5 +199,12 @@ SelectionMenu<NotificationsFilter> notificationFilterSelect(
             title: l(context).filter_read,
             icon: Symbols.mark_chat_read_rounded,
           ),
+        if (context.read<AppController>().serverSoftware ==
+            ServerSoftware.piefed)
+          SelectionMenuItem(
+            value: NotificationsFilter.read,
+            title: l(context).filter_read,
+            icon: Symbols.mark_chat_read_rounded,
+          ),
```
jwr1: Instead of adding this extra block (which looks like a duplicate of the one above), you could just change the conditional expression of the previous block, like so (which checks for either Mbin, or PieFed):

```dart
if (context.read<AppController>().serverSoftware == ServerSoftware.mbin && context.read<AppController>().serverSoftware == ServerSoftware.piefed)
```

lib/src/api/notifications.dart
jwr1, line 1: `getCount()` and `putReadAll()` are still unimplemented for PieFed. That would mean that both the notification count badge and the "Mark all as read" button would not function.

lib/src/screens/account/notification/notification_item.dart
jwr1, line 1: With the two new `NotificationType`s you added (`postMention` and `commentMention`), you'll need to update `notificationTitle` at the top of this file to include the correct corresponding messages, other wise, I believe the UI would throw an error if it tried to display those two notification types.

lib/src/api/notifications.dart
```diff
@@ -62,7 +62,13 @@ class APINotifications {
...
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path =
+            '/user/notifications/${filter == NotificationsFilter.new_ ? 'new' : (filter?.name ?? 'all')}';
+        final query = {'p': page};
+
+        final response = await client.get(path, queryParams: query);
+
+        return NotificationListModel.fromPiefed(response.bodyJson);
```
jwr1, line 76: Is paging working correctly? The way it looks right now is that it would only ever fetch one page of notifications from PieFed, as it's not correctly checking if there's a `nextPage`. You could try something like the following:

```dart
final path =  '/user/notifications/${filter == NotificationsFilter.new_ ? 'new' : (filter?.name ?? 'all')}';
final query = {'p': page};

final response = await client.get(path, queryParams: query);

final json = response.bodyJson;

return NotificationListModel.fromPiefed(json).copyWith(nextPage: lemmyCalcNextIntPage(json['items'] as List<dynamic>, page));
```

jwr1 (changes requested):
Thank you!

You working on this made me realize some of the notification code needs to be reworked a bit, but I'll do that separately.

It seems preferred.social still hasn't updated, so I won't be able to test it quite yet. Here are my initial thoughts, though.

lib/src/screens/account/notification/notification_screen.dart
JollyDevelopment: makes sense, I wasn't sure how to check for the two options so I defaulted to duping the block. will adjust to this method as its better.

Is `&&` the "or" test in flutter/dart?

lib/src/api/notifications.dart
JollyDevelopment: hmm, have not tried the mark all as read. the notification count /does/ update on the client to match the server, but not in a pattern I have been able to determine. 

I'll see if I can update that.

lib/src/screens/account/notification/notification_item.dart
JollyDevelopment: updated!

lib/src/api/notifications.dart
JollyDevelopment: ah, yeah I have not added paging on the server side yet. I will, but I have to figure out the how on that side. so far the amount of notifications in the new model way is very limited.

I can add it to my TODO list and get it in before you merge this if you prefer. otherwise I'd planned on doing that in the next PR iterrations

lib/src/screens/account/notification/notification_screen.dart
JollyDevelopment: updated!

JollyDevelopment:
note - converted back to draft to work on the `getCount()`, `putReadAll()` and paging

lib/src/screens/account/notification/notification_screen.dart
jwr1: > Is && the "or" test in flutter/dart?

OMG, I don't know why I said `&&`, it should be `||`. `&&` is **and** and `||` is **or**.

lib/src/api/notifications.dart
JollyDevelopment: i've added paging on the server side now, so this should work now. I personally have not had any more than one page of notifications so far though, so we may need to test it to be sure.

lib/src/api/notifications.dart
JollyDevelopment: I added the getCount() and putReadAll() server side handling, so I enabled those in the code here too. :-)

lib/src/api/notifications.dart
```diff
@@ -62,7 +62,16 @@ class APINotifications {
             nextPage: lemmyCalcNextIntPage(result.items, page));
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path = '/user/notifications';
+        String status = 'All'; // default to All
+        if (filter == NotificationsFilter.new_) status = 'New';
+        if (filter == NotificationsFilter.read) status = 'Read';
```
jwr1: A switch expression should be used here.

```suggestion
        final status = switch (filter) {
          NotificationsFilter.new_ => 'New',
          NotificationsFilter.read => 'Read',
          NotificationsFilter.all || null => 'All',
        };
```

lib/src/models/notification.dart
```diff
@@ -97,12 +109,25 @@ class NotificationModel with _$NotificationModel {
       creator: UserModel.fromLemmy(json['creator'] as JsonMap),
     );
   }
+
+  factory NotificationModel.fromPiefed(JsonMap json) {
+    final subject = json;
+
+    return NotificationModel(
+        id: json['notif_id'] as int,
```
jwr1, line 117: This is causing an issue while viewing the notification screen if a DM notification is present. The DM notification json returned from the API seems broken and does not contain a notification id (or most other properties for that matter). So when this line of code is ran (to obtain the notification id), an error is thrown and the notification screen crashes.

I don't mind postponing support for DM notification in the API since the DM API doesn't really work ATM either, but we do need a way to prevent it from crashing the whole notifications screen, that way people can at least use it for other types of notifications.

lib/src/api/notifications.dart
jwr1, line 1: Whenever the mark as read/unread button is used, the notification always ends up looking like it is unread. This is due to the fact that `/user/notification_state` doesn't return the notification read state in the json, and the app defaults to unread.

You can replicate this by clicking the mark as read/unread button next to a notification a few times and seeing that it looks the same no matter how many times you click on it.

jwr1 (changes requested):

lib/src/models/notification.dart
JollyDevelopment: roger, I didn't test DMs as they are not there yet in the rest of the API stuff. I'll put in a server side change to skip non-supported notifs and just not send them to the client side.

JollyDevelopment: put in a PR to skip notifs that are not properly supported yet - https://codeberg.org/rimu/pyfedi/pulls/630

lib/src/api/notifications.dart
JollyDevelopment: yep, adjusted this in this PR - https://codeberg.org/rimu/pyfedi/pulls/630 its working as intended for me in testing.

PR #176, opened 2025-05-06
Add "mark post on scroll" option.

olorin99:
This PR adds options to mark posts read on scroll. It also adds some logic to handle cases when multiple pages of read posts are encountered. When fetching a new page in the feed, if the page is empty due to all the posts being filtered out or read then it will try to read another page. If this page is also empty then it will treat it as reaching the end of recent posts and notify the user they have "caught up" on recent posts and ask if they want to try loading older posts. This is likely to only happen on less busy feeds such as in magazine feeds or the subscribed microblog feed.

2025-05-06T22:45:53,204122223+10:00


lib/src/widgets/error_page.dart
```diff
@@ -94,3 +94,62 @@ class NewPageErrorIndicator extends StatelessWidget {
...
+        },
+        child: Padding(
+          padding: const EdgeInsets.symmetric(vertical: 16),
+          child: Center(
+            child: Column(
+              mainAxisSize: MainAxisSize.min,
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                Text(
+                  'You\'re all caught up.\nLoad older items?',
```
jwr1, line 138: Could you make this a localized string?

lib/src/controller/profile.dart
jwr1, line 1: I think it would be best if we just combined `markThreadsReadOnScroll` & `markMicroblogsReadOnScroll` into one setting for all posts, just to keep it simple. Unless you had a specific reason to have separate options?

jwr1 (changes requested):
While testing, I enabled the two new settings and went back to the feed and waited for a second (expecting to see the posts on screen to be marked as read), but nothing happened. I also tried scrolling down slowly to see if that would do something, but it did not.

Am I misinterpreting what this setting does? Or am I doing something wrong?

I'll do a more thorough review once I can get it working.

lib/src/controller/profile.dart
olorin99: I've found my browsing habits are different for threads and microblogs. For microblogs I sort by newest and just scroll chronologically so marking as read on scroll is good there while for threads I'll skip over posts that I know I'll want to read later depending on mood or how much time I have to read articles etc. So for myself I have `markThreadsReadOnScroll` as disabled and `markMicroblogsReadOnScroll` as enabled.

jwr1: Ah, OK. We can leave it as is then.

lib/src/screens/feed/feed_screen.dart
```diff
@@ -999,85 +1043,144 @@ class _FeedScreenBodyState extends State<FeedScreenBody>
...
+                      return VisibilityDetector(
+                        key: Key(item.id.toString()),
+                        onVisibilityChanged: (VisibilityInfo info) {
+                          if (index <= _lastVisibleIndex &&
+                              info.visibleFraction == 0 &&
+                              _scrollDirection == ScrollDirection.reverse) {
+                            _markAsReadDebounce.run(() async {
+                              List<int> readPosts = [];
+                              for (int i = index; i >= 0; i--) {
+                                final post = _pagingController.itemList![i];
```
jwr1, line 1068: Sorry for the delay, but it does look like that fixed it. I did notice when I had read on scroll enabled and I hid read posts, that I was receiving an error if I scrolled down quite a bit and then reloaded the page:

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Null check operator used on a null value
#0      _FeedScreenBodyState.build.<anonymous closure>.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:interstellar/src/screens/feed/feed_screen.dart:1067:72)
#1      Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18:15)
#2      _Timer._runTimers (dart:isolate-patch/timer_impl.dart:410:19)
#3      _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:441:5)
#4      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:194:12)
```

It was coming from this line.

jwr1 (approved):
Thank you, it looks great!

PR #179, opened 2025-05-14
Add hero widget to advanced image.

olorin99:
Adds a hero widget to advanced image to give an animation when opening an image or navigating to an image post.
The 'hero' parameter in AdvancedImage is so cross posted images have an unique tag otherwise they don't animate properly when multiple of the same image are shown in a feed.

### DIFF
--- a/lib/src/widgets/image.dart
+++ b/lib/src/widgets/image.dart
@@ -45,7 +45,8 @@ class AdvancedImage extends StatelessWidget {
                 pageBuilder: (context, _, __) => AdvancedImagePage(
                   image,
                   title: openTitle!,
-                  hero: hero
+                  hero: hero,
+                  fit: fit,
                 ),
               ),
             );
@@ -87,8 +88,10 @@ class AdvancedImagePage extends StatefulWidget {
   final ImageModel image;
   final String title;
   final String? hero;
+  final BoxFit fit;
 
-  const AdvancedImagePage(this.image, {super.key, required this.title, this.hero});
+  const AdvancedImagePage(this.image,
+      {super.key, required this.title, this.hero, this.fit = BoxFit.contain});
 
   @override
   State<AdvancedImagePage> createState() => _AdvancedImagePageState();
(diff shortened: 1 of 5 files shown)


lib/src/widgets/image.dart
```diff
@@ -31,56 +33,62 @@ class AdvancedImage extends StatelessWidget {
...
-                      (blurHashSizeFactor! * image.blurHashWidth!).ceil(),
-                  decodingHeight:
-                      (blurHashSizeFactor * image.blurHashHeight!).ceil(),
-                  scale: blurHashSizeFactor,
+        shouldWrap: openTitle != null,
+        parentBuilder: (child) => GestureDetector(
+          onTap: () {
+            Navigator.of(context).push(
+              PageRouteBuilder(
+                transitionDuration: Duration(seconds: 1),
```
jwr1: The 1 second duration seemed like it was _way_ too slow. I changed it locally to 500 milliseconds and that seemed a lot better, but then I realized the default duration is 300 milliseconds, so maybe we should just stick with that instead.

```suggestion
```

lib/src/widgets/image.dart
jwr1, line 1: I noticed that the animation has sizing issues immediately once starting the animation, both when opening an image and when closing the image. In the following example, I used `timeDilation = 10;` in `AdvancedImage.build` to make the problem obvious. You can see that when I click on the image, it instantly shrinks once the animation starts, and when closing out the image, it instantly blows up once the animation starts. The issue seems to very based on the aspect ratio of the image and how it compares to the app window. I'm not really sure if this is something we can fix on our end, but I thought it was worth a shot, cause the issue is definitely noticeable on certain images.

[Screencast_20250514_105257.webm](https://github.com/user-attachments/assets/f91c583f-337f-49db-932d-d2b608fb9bb9)

jwr1 (changes requested):
Thanks, I actually did look into using `Hero` over a year ago (although it was for opening the whole post, not the image), so I'm glad you were able to loop back to this.

lib/src/widgets/image.dart
olorin99: I just did some testing and it looks like this is only an issue when the fit is `BoxFit.cover`. Full size images seem to transition as expected. Not sure how to fix apart from changing the fit and that would make non full size images not fit within the layout. I'll look into it.

olorin99:
Would you like me to add another hero animation for whole posts as well as images?

jwr1:
> Would you like me to add another hero animation for whole posts as well as images?

Sure, might as well!

olorin99:
That should fix the animation issues. The problem was flutter can't animate between different BoxFits so keeping the same fit between image and image page and using Center to keep the image contained should solve the problem.

jwr1 (approved):
Awesome, it looks great!

PR #182, opened 2025-05-17
Add blur to markdown embedded images if post is nsfw.

olorin99:
Noticed some posts that were marked as nsfw but the embedded images weren't blurred only the attached image was. This should fix that. Happily it wasn't anything super weird, some people just mark everything as nsfw.


jwr1 (approved):
LGTM!

PR #184, opened 2025-05-20
Update to Dart 3.8

jwr1:
- Bump Dart and Flutter versions.
- Use the new [null-aware elements](https://dart.dev/language/collections#null-aware-element) from Dart 3.8. I was able to find 13 places to use it, but I'm sure there's more hiding elsewhere.
- Run `dart format .` with the Dart 3.8 formatter changes.
- Minor bump of dependency versions. There are several dependencies with major updates, but those will require more work and can be done separately.


olorin99 (approved):
Looks good. The null-aware elements are pretty nice.

Issue #173, opened 2025-05-05
(osx-x86) even with compact mode set, there's a LOT of wasted screen space/big fonts


akohlsmith:
### Describe the bug

See the attached screenshot. The "sub/mod/fav/all" tabs have a LOT of vertical space. It feels like they should be no taller than regular browser tabs, especially in compact mode.

Perhaps adding a font size slider like many other apps would be a good solution to controlling the overall size of text as well? Right now it feels a little like I'm reading using a screen magnifier. :-)

<img width="877" alt="Image" src="https://github.com/user-attachments/assets/888b775b-b97c-4d57-a61c-bb3ac5302efa" />

### Version

0.9.1 (1)

### What client platform(s) are you seeing the problem on?

Linux

### What server platform(s) are you seeing the problem on?

Mbin

### Additional context

Just one other minor gotcha -- the "What client platform(s) are you seeing the problem on" question needs dropdown selections for osx and ios as well now

akohlsmith:
The same gigantic whitespace is present on iOS (iphone 13 mini)

jwr1:
I know this isn't a solution to the issue, but you can always hide those tabs in Feed. You just need to change the Feed Actions settings.

jwr1:
I've experimented with changing both the `textScaler` value and the `fontSize`s being passed down in the `Theme`, and they both proved to be ineffective with UI spacing.

<details>
<summary>Examples</summary>

Here's `textScaler = 0.5`:

![Image](https://github.com/user-attachments/assets/c52c9683-3f82-4752-ab51-c09b884e78a9)

Here's `textScaler = 1`:

![Image](https://github.com/user-attachments/assets/d3ab42f7-51d5-4a2f-ab7f-c7f9e928ad9b)

</details>

As you can see in the examples, only the actual text size is changed, not the UI spacing. You can try out the text size changes in https://github.com/interstellar-app/interstellar/pull/187. I also tried setting `visualDensity` on `Theme`, but it didn't seem like it was changing anything.

Not sure if you have any other ideas, @olorin99?

Issue #107, opened 2025-02-26
UI UX tweaks


breadguyyy:
### Description

- default empty magazine field to random
- give magazine and user mentions colored text and bg for visibility (like misskey for example)
- color gradient for thread comment bands
- fetch remote follows/followers on user profiles
- when tapping the reply button, automatically open keyboard and jump to where the reply field is visible above it
- add the reply button next to the comment button in feed
- imgur embeds?

jwr1:
> * give magazine and user mentions colored text and bg for visibility (like misskey for example)

Would you have an example of what misskey's mentions look like?

> * fetch remote follows/followers on user profiles

Not sure what you mean here. You can see how many followers a user has by looking at the Follow button. You can see who follows a user by looking at the Followers tab and you can see who the user is following by looking at the Following tab.

> * add the reply button next to the comment button in feed

There's no reply button here because the majority of the threads' bodies are cut off (limited to 4 lines of text), so you won't get the whole context till you click on it.

> * image embeds?

With Mbin's API, there is no way to directly upload an image and then embed it somewhere in a post/comment. What you have to do is upload it as an attachment, which Interstellar dosen't let you do for comments yet.
 
> * I can't remember what the current defaults are but definitely have set filter on the app bar, and default sort by hot for everything.

I'd like to at least have something used for the tabs by default; otherwise, people won't even know it's possible without digging through the settings. We could either leave it to the "Set filter" (Sub, Mod, Fav, All) option, or change it to the "Set type" (Threads, Microblogs) option.

breadguyyy:
here's a cherrypick (mk fork) mention example

![Image](https://github.com/user-attachments/assets/0301b0f9-7e28-48e7-80a4-5f7dafdb8fab)

breadguyyy:
also for fetching remote follows I mean that it currently only shows followers/follows that are already federated

![Image](https://github.com/user-attachments/assets/c6415776-41d1-4ea1-b920-7a0775179932)

breadguyyy:
I shouldn't have split all these replies out but whatever lol. fair enough about the tabs, they'll probably be fine if/when you add an option to hide the toolbars while scrolling

jwr1:
> also for fetching remote follows I mean that it currently only shows followers/follows that are already federated
> 
> ![Image](https://github.com/user-attachments/assets/c6415776-41d1-4ea1-b920-7a0775179932)

Yeah, that would be something Mbin would need to fix.

[...]

PR #186, opened 2025-05-26
Hide ui elements on scroll

olorin99:
Adds options to hide the appbar, tabs, and or FAB on scrolling in the feed.
Partially addresses #173 but should probably add a text size slider for desktop as well. Also addresses a listed item in #107.

### DIFF
--- a/lib/src/screens/settings/display_screen.dart
+++ b/lib/src/screens/settings/display_screen.dart
@@ -84,6 +84,7 @@ class DisplaySettingsScreen extends StatelessWidget {
                     ac.selectedProfileValue.copyWith(enableTrueBlack: newValue),
                   ),
           ),
+          const Divider(),
           ListTileSwitch(
             leading: const Icon(Symbols.view_agenda_rounded),
             title: Text(l(context).settings_compactMode),


jwr1 (changes requested):
Looks nice! Just have a few things:

- Similar to last time, I'm not sure we really need three separate settings for the three different UI elements that could hide on scroll (unless you specifically had a reason). I would combine it into one setting like "Hide feed UI on scroll".
- What exactly is `SubordinateScrollController` for? Maybe you could add a comment to the class with a brief explanation on what it does/is used for.
- If you scroll on Feed so that the UI is hidden, then switch to a different screen (e.g., Settings), then switch back to Feed and scroll up, a bunch of errors will be thrown.
- If you scroll on Feed so that the UI is hidden, then swipe to the right to view the next tab from Feed (such as the Mod view), then if you try scrolling down, nothing will happen. You have to first scroll up then down in order for it to work.

olorin99:
- Just went with more customization = good. But easy to simplify it.
- A `ScrollController` can only have one scroll position attached to it at a time and when using tabs multiple `CustomScrollViews` would try and attach to the `NestedScrollView`. The `SubordinateScrollController` attaches or detaches the scroll position whenever its "active status" changes (i.e. whenever the tab changes).
- Should be fixed now.
- Should be fixed now.

PR #191, opened 2025-05-30
Piefed - Community Moderation

JollyDevelopment:
This adds the ability for users to see the list of banned users for a community the user moderates or owns. 

This also enables the 'unban' function for those banned users.

This also enables the 'ban' function so community mods can ban a user from their community.

This also enables the 'markNsfw' function so community mods can update a post status.

piefed side PR: https://codeberg.org/rimu/pyfedi/pulls/675

### DIFF
--- a/lib/src/api/magazine_moderation.dart
+++ b/lib/src/api/magazine_moderation.dart
@@ -56,7 +56,17 @@ class APIMagazineModeration {
         throw Exception('Ban update not implemented on Lemmy yet');
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path = '/community/moderate/ban';
+        final body = {
+          'community_id': magazineId,
+          'user_id': userId,
+          'reason': reason,
+          'expiredAt': expiredAt?.toIso8601String(),
+        };
+
+        final response = await client.post(path, body: body);
+
+        return MagazineBanModel.fromPiefed(response.bodyJson);
     }
   }
 
--- a/lib/src/api/magazine_moderation.dart
+++ b/lib/src/api/magazine_moderation.dart
@@ -26,7 +26,7 @@ class APIMagazineModeration {
 
         final body = {
           'community_id': magazineId,
-          'page': (page == null) ? 1 : page,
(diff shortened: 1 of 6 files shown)


lib/src/api/magazine_moderation.dart
```diff
@@ -22,7 +22,16 @@ class APIMagazineModeration {
         throw Exception('List banned users not allowed on lemmy');
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path = '/community/moderate/bans';
+
+        final body = {
+          'community_id': magazineId,
+          'page': (page == null) ? 1 : page,
```
jwr1: The `??` operator checks if _left_ is not `null` and uses it if that's true, otherwise, if it is `null`, it falls back to _right_.

```suggestion
          'page': page ?? 1,
```

Does this API endpoint actually require a `page` though? Most of the endpoints default to `1` if `page` is not provided.

JollyDevelopment: I just based it on the mbin versions really. my test instance and user only has like two bans for the test community, but I want to accommodate for bigger communities that will have much bigger lists.

jwr1: Oh, I wasn't saying it shouldn't be paged. I think it should.

I was just asking if the `page` property was optional in the request body (where it would default to `1` if not provided)?

JollyDevelopment: yeah it does default to `1` if `page` is not in the request, but i kept getting errors when sending the initial call from interstellar when `page` was null, which translated to `None` on the server side.

the firstPageKey starts as `''` here: https://github.com/JollyDevelopment/interstellar/blob/jollydev/piefed-community-moderation-list-bans-unban/lib/src/screens/explore/magazine_mod_panel.dart#L78 

so I figured I could check if its the first call and if so /not/ send the `page` in the body, or just send `page:1` if its the first call. 

one of those half-dozen-of-one six-of-another sort of calls.

jwr1 (approved):
Thanks, LGTM!

PR #195, opened 2025-06-04
Add tab for viewing and responding to reports in mbin.

### LINKED ISSUE #36: Moderation support
### Describe the feature you'd like to request

- [ ] View magazine rules
- [ ] View magazine owner and moderators
- [ ] Edit magazine details
- [ ] Mod actions (pin, remove, etc.) on content
- [ ] Create magazine
- [ ] View and manage reports

### Additional context

_No response_

olorin99:
Adds tab to mod panel for viewing and responding to reports.
Closes #36.


lib/src/api/community_moderation.dart
```diff
@@ -4,11 +4,66 @@ import 'package:interstellar/src/models/community.dart';
 import 'package:interstellar/src/utils/models.dart';
 import 'package:interstellar/src/utils/utils.dart';
 
+enum ReportStatus { any, appeal, approved, closed, pending, rejected }
```
jwr1: I haven't even heard of some of these report types while using Mbin. I know you can approve and reject reports with the Mbin web UI, but I don't think there's a way to actually appeal a report is there? Correct me if I'm wrong, but I feel like we should be able to simplify this by ignoring the `appeal` and `closed` status. The Mbin web UI, for example, only has filters for pending, approved, rejected, and all.

lib/src/screens/explore/community_mod_panel.dart
```diff
@@ -42,16 +51,29 @@ class _CommunityModPanelState extends State<CommunityModPanel> {
     }
 
     return DefaultTabController(
-      length: 1,
+      length:
+          context.read<AppController>().serverSoftware == ServerSoftware.mbin
+          ? 2
+          : 1,
```
jwr1, line 58: You can remove this check since the Mod Panel is only accessible on Mbin. Once support for Lemmy & PieFed is added for the Mod Panel, we can also add support for viewing reports.

lib/src/screens/explore/community_mod_panel.dart
```diff
@@ -42,16 +51,29 @@ class _CommunityModPanelState extends State<CommunityModPanel> {
...
       child: Scaffold(
         appBar: AppBar(
           title: Text('Mod Panel for ${widget.initData.name}'),
-          bottom: const TabBar(tabs: <Widget>[Tab(text: 'Bans')]),
+          bottom: TabBar(
+            tabs: [
+              Tab(text: 'Bans'),
+              if (context.read<AppController>().serverSoftware ==
+                  ServerSoftware.mbin)
```
jwr1, line 66: > You can remove this check since the Mod Panel is only accessible on Mbin.

Same here.

lib/src/screens/explore/community_mod_panel.dart
```diff
@@ -42,16 +51,29 @@ class _CommunityModPanelState extends State<CommunityModPanel> {
...
+                Tab(text: 'Reports'),
+            ],
+          ),
         ),
         body: TabBarView(
           physics: appTabViewPhysics(context),
           children: <Widget>[
             CommunityModPanelBans(data: _data, onUpdate: onUpdate),
+            if (context.read<AppController>().serverSoftware ==
+                ServerSoftware.mbin)
```
jwr1, line 76: > You can remove this check since the Mod Panel is only accessible on Mbin.

And here.

lib/src/models/community.dart
```diff
@@ -254,3 +256,75 @@ class CommunityBanModel with _$CommunityBanModel {
...
+class CommunityReportModel with _$CommunityReportModel {
+  const factory CommunityReportModel({
+    required int id,
+    required CommunityModel? community,
+    required UserModel? reportedBy,
+    required UserModel? reportedUser,
+    required PostModel? subjectPost,
+    required CommentModel? subjectComment,
+    required String? reason,
+    required String? status,
```
jwr1: Can this use the `ReportStatus` enum?

lib/src/screens/explore/community_mod_panel.dart
```diff
@@ -165,3 +187,324 @@ class _CommunityModPanelBansState extends State<CommunityModPanelBans> {
     super.dispose();
   }
 }
+
+class CommunityModPanelReports extends StatefulWidget {
```
jwr1, line 192: I just have a few suggestions for the UI here, as I'd say things seem a bit cluttered/verbose at the moment.

- Can the View button be removed and replaced with an InkWell that opens the reported content when the report is clicked on?
- Can the Ban User button be changed to include the actual user's username, similar to how Mbin does it? Then from here, we can also remove the "Reported user" text, as you can see which user is being reported in the Ban X button.
- Maybe to simplify the actions, hide the reject button if the report is already rejected, hide the accept button if already approved, and display both if still pending; that's what Mbin does for reports as well.

Let's see how it looks after your next changes and go from there.

jwr1 (changes requested):
I finally got around to reviewing this, so here we go :)

lib/src/screens/explore/community_mod_panel.dart
olorin99: Seems the mod panel is also accessible from PieFed so I think these checks are still needed.

jwr1: Hm, well I can't access it from my piefed.social account for a community I moderate. I believe the only way it's accessible is if the list of mods from a community includes your signed-in username. On PieFed and Lemmy, we don't even correctly bring in the list of moderators, so the list is always empty, and the mod panel is never accessible.

olorin99: Weird I can access it from the test community I made on preferred.social

olorin99: Looks like it depends on where the community info is fetched from. `/community` includes moderator info while `/community/list` doesn't. So opening from the feed or a post will show mods but opening from search will not. Looks like the simplest fix is to always call `api.community.get()` when opening a community rather than only when `_data == null`.

jwr1: Ah, ok. We can leave it then.

jwr1 (approved):
Turns out Mbin does still use the `closed` report status, even though it should be `approved`. I'm going to create an issue on the Mbin repo for that, but I did create a temp fix here as well.

I also just fix it so text wrapping would work (for small screens or long report reasons), and also made it so the flex direction of the report actions would change to horizontal if there's enough screen space.

Thanks!

jwr1:
I've just created an issue on Mbin about those extra report status types: https://github.com/MbinOrg/mbin/issues/1639

PR #218, opened 2025-06-14
Add page to view logged in users bookmarks.

olorin99:
Adds a page to view the logged in users bookmarks.
The page is accessible from the profile screen as a icon button in the appbar, left of the sort.
There doesn't seem to be support yet for fetching saved posts on piefed but should work for lemmy and mbin.


lib/src/api/bookmark.dart
jwr1, line 1: When I view my bookmarks on Lemmy, once the first page is loaded, it seems like it's trying to load the next page forever. Is it possible `next_page` isn't being set/handled correctly?

jwr1, line 1: I noticed with Lemmy that the bookmarks don't display any saved comments (only posts). Wasn't sure if this was intentional or not?

jwr1, line 1: It seems no matter which bookmark list I view on Mbin, it will always display the contents of the default list.

lib/src/screens/explore/bookmarks_screen.dart
jwr1, line 1: Maybe we should have a dialog to confirm deletion, because I could see someone accidentally deleting a bookmark list, considering how easy it is to click.

lib/src/api/bookmark.dart
olorin99: It seems the pagingcontroller struggles when using a page_cursor from lemmy. When using a page_cursor it takes a while before triggering the next `_fetchPage` call. I switched to using a page number and it triggered the call immediately. Same number of api calls for each, both worked, just using the page_cursor took longer for some reason. I've left it using the page number.

olorin99: Ahh mbin fetches posts and comments together and lemmy fetches them separately. Forgot to fetch comments too.

olorin99: I think this is a bug on the mbin side. I've added an issue on the mbin repo https://github.com/MbinOrg/mbin/issues/1644.

jwr1 (approved):

PR #221, opened 2025-06-23
Feeds

olorin99:
Adds user created custom feeds.
- Feeds can be composed of multiple communities or users.
- Feeds can be shared between users like profiles or filter lists.
- If an input isn't federated to the current logged in instance the input is ignored.

Regarding future PieFed feeds support.
An Interstellar feed can technically be composed of any combination of FeedSource (all, subscribed, user, community, etc...) but currently only communities and users can be added by the user. To work we just need a method to list posts from the source and another to get the source by name. I expect the PieFed feeds API to be similar to that of users and communities so it should be no problem to be able to include PieFed feeds as inputs into Interstellar feeds.


lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -0,0 +1,700 @@
...
+    setState(() {
+      feedData = feedData.copyWith(inputs: inputs);
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final ac = context.watch<AppController>();
+    return Scaffold(
+      appBar: AppBar(title: Text(l(context).feeds_edit(widget.feed!))),
```
jwr1: Hmm, I was going to try to test this PR, but when I try to add a new feed, the app completely crashes.

```
Null check operator used on a null value
```

jwr1:
I can't tell that this is actually working. I created a Feed and added piracy@lemmy.dbzer0.com and technology@lemmy.world, both quite popular communities. However, when I actually view the feed, all I'm seeing are posts from technology.

olorin99:
It looks like pinned posts were getting in the way of the merging. It should work now.

lib/src/screens/settings/settings_screen.dart
```diff
@@ -45,6 +46,15 @@ class SettingsScreen extends StatelessWidget {
               ),
             ),
           ),
+          ListTile(
+            leading: const Icon(Symbols.feed_rounded),
+            title: Text('Feeds'),
```
jwr1: This should be changed to a localized string.

lib/src/screens/feed/feed_screen.dart
jwr1, line 1: When a feed is being viewed, I think it would be good to have the app bar title set to the name of the feed.

lib/src/screens/feed/nav_drawer.dart
jwr1, line 1: I know this would be annoying to do, but can we make the nav dropdown state persist (between startup)? It doesn't even need to be in the `profile`, just its own separate state.

I just think most people would use a few of the dropdowns and not the rest, and would be annoyed at having to expand the same dropdowns repeatedly.

jwr1, line 1: Maybe it's just me, but I feel like the expand/collapse tiles are just a tad taller than they should be. They just seem to take up a lot of space.

lib/src/screens/feed/feed_screen.dart
```diff
@@ -551,13 +565,15 @@ class _FeedScreenState extends State<FeedScreen>
...
+        drawer: (widget.sourceId != null || widget.feed != null)
+            ? null
+            : const NavDrawer(),
       ),
     );
   }
 }
 
-enum FeedView { threads, microblog, timeline }
+enum FeedView { threads, microblog, combined }
```
jwr1: Unfortunately, the problem with renaming this enum value verbatim is that it will break (and crash) all the `fromJson` methods if a client was previously set to `timeline`. We will need some way to migrate (or ignore) the old `timeline` value. I wonder if json_seriazliable would let us set a custom `fromJson` function on an enum.

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -0,0 +1,700 @@
...
+    final currentExploreSort = exploreSortSelection(context).getOption(sort);
+    final currentExploreFilter = exploreFilterSelection(
+      context,
+      type,
+    ).getOption(filter);
+
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(
+          '${l(context).explore} ${context.watch<AppController>().instanceHost}',
```
jwr1: I feel like this should be updated to reflect you're editing the feed, more specifically, adding inputs.

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -0,0 +1,700 @@
...
+                label: Text(l(context).feeds_delete),
+              ),
+            ),
+        ],
+      ),
+    );
+  }
+}
+
+class ExploreFeedInputsScreen extends StatefulWidget {
```
jwr1: Is there any possible way to dedpulicate this with the explore screen? I just hate that if we updated something on the explore page we'd have to remember to come back here and update things as well.

jwr1: Could we have this class in a separate file?

jwr1: I feel like we might want to rework the UI for the selection input tiles just a bit. The primary thing we're doing on this screen is adding inputs to a feed, so when a community/user/domain is clicked on, it seems like that should check the box and add it as an input (not actually open it). Additionally, we don't really need the option to subscribe here, so it seems like we could remove that button so it's out of the way.

In my mind, I think the best way to have it is on the right side of a tile is a checkbox and an open button. Clicking either the tile itself or the checkbox will enable the input, and clicking the open button will actually open it so you can quickly view it. Kind of like what the community selection input looks like, but with a checkbox where the red circle is.



I know it won't look perfect with these changes, but I feel like it would at least be an improvement. What do you think?

olorin99: I've updated the explore screen with the extra select inputs functionality and removed the extra class.

lib/src/screens/feed/feed_screen.dart
olorin99: Didn't think about this. I reverted the commit and changed so only the ui is changed. I'll look for a way to migrate between saved values in a different PR.

lib/src/screens/feed/nav_drawer.dart
olorin99: Added a store for misc values we want to cache. Also ended up caching the rest of the nav drawer state in feed screen.

jwr1, line 1: The nav cache isn't working for me. I opened a few of the dropdowns and reloaded the app, and all the dropdowns were collapsed again. It seems like it is saving the cache value correctly (I can see it in storage), it's just not initializing/applying the state correctly, I guess.

lib/src/screens/feed/feed_screen.dart
```diff
@@ -93,6 +99,14 @@ class _FeedScreenState extends State<FeedScreen>
         .read<AppController>()
         .profile
         .feedDefaultHideReadPosts;
+
+    _navDrawPersistentState = NavDrawPersistentState();
+    () async {
+      final drawerState = await fetchNavDrawerState(context.read<AppController>());
+      setState(() {
+        _navDrawPersistentState = drawerState;
+      });
```
jwr1: I got a state error here.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: setState() called after dispose(): _FeedScreenState#00fdc(lifecycle state: defunct, not mounted)
This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.
The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().
#0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1163:9)
#1      State.setState (package:flutter/src/widgets/framework.dart:1198:6)
#2      _FeedScreenState.initState.<anonymous closure> (package:interstellar/src/screens/feed/feed_screen.dart:106:7)
<asynchronous suspension>
```

lib/src/screens/feed/feed_screen.dart
```diff
@@ -563,7 +589,18 @@ class _FeedScreenState extends State<FeedScreen>
...
+        drawer: (widget.sourceId != null || widget.feed != null)
+            ? null
+            : NavDrawer(
+              drawerState: _navDrawPersistentState,
+              updateState: (NavDrawPersistentState? drawerState) async {
+                drawerState ??= await fetchNavDrawerState(context.read<AppController>());
+
+                setState(() {
+                  _navDrawPersistentState = drawerState!;
+                });
```
jwr1, line 594: I got a state error here.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: setState() called after dispose(): _FeedScreenState#00fdc(lifecycle state: defunct, not mounted)
This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.
The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().
#0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1163:9)
#1      State.setState (package:flutter/src/widgets/framework.dart:1198:6)
#2      _FeedScreenState.build.<anonymous closure> (package:interstellar/src/screens/feed/feed_screen.dart:599:17)
<asynchronous suspension>
```

jwr1 (changes requested):

lib/src/screens/feed/nav_drawer.dart
jwr1, line 1: Does the nav cache for community, user, and domain subscriptions account for the user account changing? (Does it invalidate?)

lib/src/screens/explore/explore_screen.dart
jwr1, line 1: The "All" tab is now showing when adding a feed input.

jwr1, line 1: I'm not sure that the domain inputs are working. I added `kbin.earth` as an input, and then it's showing `kbin.earth@kbin.earth` as the input on the feed overview screen (should it just be `kbin.earth` shown here?). Then when I actually go to view the feed, there is nothing shown, even though if I view the `kbin.earth` domain separately, it does show threads.

lib/src/screens/feed/nav_drawer.dart
olorin99: The initial expansion requires valid values on the first build. This works fine on a narrow screen since the nav drawer is recreated each time its accessed but on a wide screen its persistent so I changed it so it shows a loading indicator until the values are available so the dropdowns should now init expanded properly.

olorin99: Yeah, the subscriptions are cached in feed screen and since feed screen is invalidated whenever an account or profile is changed the cache will also be invalidated. Only the expanded status for the dropdowns is cached independently.

lib/src/screens/explore/explore_screen.dart
olorin99: Yeah domains don't work. The way I've got it currently is that the input info is stored as a string (interstellar@kbin.earth, olorin99@kbin.earth, etc) and then fetched via the corresponding `getByName` API call. Since domains don't have a call to get by name, they can only be gotten by id, it wouldn't work across instances.

I was thinking about adding a cache for ids which would also help when accessing larger feeds. This would mean though that when accessing a feed for the first time on a different instance it would have to loop through all the values returned by `/api/domains` to find the id that matches the domain. Which depending on the domain might take a while.

Alternatively we can add an feature request for mbin to add a `/api/domain/name/{domain_name}` call similar to the ones for magazines and users.

In the meantime I'll just hide the domains from input search.

jwr1: Honestly, I don't really care about domains, so I wouldn't mind scrapping it, at least for trying to use it as a Feed input.

jwr1 (approved):
Thanks, looks great!

PR #223, opened 2025-06-24
Add setting to adjust the app animation speed.

### LINKED ISSUE #201: Possibly add a way to speed up/disable UI animations
### Describe the feature

Hey there!

I really like the new animations present in the app, but it'd be nice to be able to speed them up (or disable them if speeding them up isn't possible). I'm imagining a UI slider that lets you set the animation speeds to 1x-4x, with the furthest right position being a "disable" toggle.

### Additional context

_No response_

olorin99:
Adds settings to scale animation speed between 0 (off) to 4x the default speed.
Adjusts the duration for post and image hero animations as well as for the floating menu animations. Is there anywhere else that needs it?

Closes #201

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -194,6 +194,8 @@
   "settings_autoTranslate": "Auto translate posts/comments",
   "settings_markThreadsReadOnScroll": "Mark threads read on scroll",
   "settings_markMicroblogsReadOnScroll": "Mark microblogs read on scroll",
+  "settings_animationSpeed": "Animation speed",
+  "settings_animationDisabled": "Disabled",
   "settings_themeMode": "Theme mode",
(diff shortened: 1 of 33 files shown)


jwr1 (changes requested):
I can think of two more places we'd want this setting to affect:

- There's the `themeAnimationDuration` property on `MaterialApp`, which is used for theme change animations.
- There's also all the other page transitions, like if you tap on a community, you can see there's a little zoom in effect, and zoom out for navigating back. We definitely don't want to have to manually pass in a transition duration for every single route change in the app, so we just need to figure out the best way to pass it in.

olorin99:
Added a utility function for pushing routes so if we ever want to adjust every route in the app again we should only need to adjust one place. Hopefully I got every route change.

jwr1 (approved):

PR #225, opened 2025-07-07
Assorted minor fixes.

olorin99:
Mostly changing some `if (!mounted)` to `if (!context.mounted)` as suggested by dart analysis.

Added check for `direction != _scrollDirection` in feed screen getScrollDirection. Without this it was triggering rebuilds whenever the screen was scrolled. I can't really see a difference but there was a user who had stuttering when they scrolled so maybe this will help a little.

Also wrapped content item in a `RepaintBoundary`. This should prevent changes in the wrapped item from causing repaints in the rest of the feed.

Added a null check and a check for mounted context to post items onUpdate callback. Found a null access bug when voting on a post then quickly exiting the page.

### DIFF
--- a/lib/src/controller/controller.dart
+++ b/lib/src/controller/controller.dart
@@ -475,6 +475,7 @@ class AppController with ChangeNotifier {
       throw Exception('Notification permissions denied');
     }
 
+    if (!context.mounted) return;
     await UnifiedPush.registerAppWithDialog(context, _selectedAccount, [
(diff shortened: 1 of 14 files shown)


olorin99:
I believe `mounted` is for guarding state and `context.mounted` is for guarding the build context. Not entirely sure when which one is need, I'm just following the IDE suggestions :).

jwr1 (approved):

PR #226, opened 2025-07-07
Use piefeds api to mark as read when available.

olorin99:


jwr1 (approved):
The Post model wasn't using PieFed's read property, but I went ahead and fixed that. Other than that, it looks good.

PR #228, opened 2025-07-07
Add secondary actions to the navigation bar.

### LINKED ISSUE #215: QoL: Tap feed button to return to top
### Describe the feature

Tapping the feed button while in the feed view would be a nice way to return to the top, the floating hamburger menu to the right of the screen currently makes it a two handed operation on a large phone.

### Additional context

_No response_

olorin99:
When tapping feed navigation destination while the feed is open it scrolls to the top.
Tapping explore while explore page is open will move focus to the search bar.
Tapping account while account page is open will open the account switcher.
Closes #215


jwr1 (changes requested):
Since the Account nav switches accounts, maybe we could have the Settings nav switch profiles (since it doesn't do anything right now). Besides that, I'd say it's looking good.

PR #231, opened 2025-07-14
Handle back button navigation on main screen.

olorin99:
Pressing back will return to main feed screen from explore/account/settings.
Pressing back twice from main feed will exit app.


jwr1 (changes requested):
Not sure if there's any way to fix this, but if I open Explore, Account, or Settings, and then use system back to go back to feed, the indicators in the nav look messed up.

jwr1 (approved):
Ok, then I'll go ahead a merge. Maybe I could try to figure it out later.

PR #235, opened 2025-07-18
New chat button.

### LINKED ISSUE #135: new chat button
### Describe the feature

add a button on the messages page to search a username and send a message, just as an alternative to looking up their username from the explore page

### Additional context

_No response_

olorin99:
- Add piefed send message api.
- Adds new chat floating button on messages screen. Closes #135 
- Change user screen send message from inline to open message thread screen.

On PieFed and Lemmy when navigating to a message thread from the user screen it should fill with the previous messages with the user however on Mbin there isn't a way to map a user id to a message thread id so can only get previous messages when navigating from the messages screen.

### DIFF
--- a/lib/src/api/messages.dart
+++ b/lib/src/api/messages.dart
@@ -135,7 +135,21 @@ class APIMessages {
         ).items.first;
 
       case ServerSoftware.piefed:
-        throw UnimplementedError();
+        final path = '/private_message';
+
+        final response = await client.post(
+            path,
+            body: {'recipient_id': userId, 'content': body}
+        );
+
+        final json = response.bodyJson;
+
+        return MessageListModel.fromPiefed(
+          {
+            'private_messages': [json['private_message_view']],
+          },
+          (json as dynamic)['private_message_view']['creator']['id'] as int,
+        ).items.first;
     }
   }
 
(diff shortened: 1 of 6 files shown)


jwr1 (changes requested):
Looks good, just a few things:

- The title for the new chat page shouldn't be "User Follows".
- On the new chat page, the "People" option is visible, albeit the only option, but it probably shouldn't be there at all, considering you can't message any other explore type.
- On Mbin, if I create a new chat after searching a user, and then send a message, if I send more than one message at once (while still on that screen), it will create multiple new chats, instead of using the same one. I believe what we need to do here is to use the `create` message API on the first send, and then `postThreadReply` on all subsequent messages sent on that screen.

Just a note, I'm not exactly sure what the 403 errors you were running into earlier were caused by, but could it be due to an existing message thread already existing with an account?

jwr1 (changes requested):
Ok, just one more thing to add. If I create a new chat, and then go back to the messages screen, then the new chat doesn't show up yet until the message list is reloaded, which is a problem considering how difficult it is to get it reloaded on desktop. On mobile, you would just pull to refresh, but that doesn't work on desktop. Before the screen state caching was added, you'd just be able to switch to a different tab and then switch back, but that obviously doesn't work anymore.

Anyway, how about if a new chat is created, then the message list screen should just be refreshed automatically?

jwr1 (approved):

PR #237, opened 2025-07-21
Fix saving piefed user cover.

### LINKED ISSUE #213: User avatar and cover editing with PieFed
### Description

Ensure user avatar and cover editing work with PieFed. There's already code that exists for the API, but it does need to be verified as working.

olorin99:
Tested both avatar and cover on piefed.social.
Closes #213


jwr1 (approved):
Thanks!

PR #238, opened 2025-07-21
Debug settings

### LINKED ISSUE #229: disable errors
### Describe the feature

Add an option to disable error messages

### Additional context

_No response_

olorin99:
- Add option to show/hide error pop ups.
- Add simple logger. 
    - The output can be seen under Debug > Log.
    - Individual lines can be copied to clipboard or the whole log can be saved to a file.
- Add options to clear database/accounts/profiles.

Closes #229

### DIFF
--- a/lib/src/controller/profile.dart
+++ b/lib/src/controller/profile.dart
@@ -229,7 +229,7 @@ class ProfileRequired with _$ProfileRequired {
     swipeActionRightLong: SwipeAction.reply,
     swipeActionThreshold: 0.20,
     filterLists: {},
-    showErrors: false,
+    showErrors: true,
   );
 }
 
--- a/lib/src/screens/settings/settings_screen.dart
+++ b/lib/src/screens/settings/settings_screen.dart
@@ -5,7 +5,6 @@ import 'package:interstellar/src/screens/settings/about_screen.dart';
 import 'package:interstellar/src/screens/settings/account_selection.dart';
 import 'package:interstellar/src/screens/settings/behavior_screen.dart';
 import 'package:interstellar/src/screens/settings/data_utilities.dart';
-import 'package:interstellar/src/screens/settings/debug/debug_screen.dart';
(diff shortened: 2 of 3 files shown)


lib/src/screens/settings/settings_screen.dart
```diff
@@ -101,6 +102,14 @@ class SettingsScreen extends StatelessWidget {
               builder: (context) => const DataUtilitiesScreen(),
             ),
           ),
+          ListTile(
+            leading: const Icon(Symbols.developer_board_rounded),
```
jwr1: The bug icon is usually used for debug menus.

```suggestion
            leading: const Icon(Symbols.bug_report_rounded),
```

lib/src/screens/settings/settings_screen.dart
```diff
@@ -101,6 +102,14 @@ class SettingsScreen extends StatelessWidget {
...
+            title: Text(l(context).settings_debug),
+            onTap: () => pushRoute(
+              context,
+              builder: (context) => const DebugSettingsScreen()
+            ),
+          ),
```
jwr1: Could we move this to the About screen, maybe right at the top? I'm guessing most people won't be accessing this screen, at least not frequently.

lib/src/controller/profile.dart
```diff
@@ -227,6 +229,7 @@ class ProfileRequired with _$ProfileRequired {
     swipeActionRightLong: SwipeAction.reply,
     swipeActionThreshold: 0.20,
     filterLists: {},
+    showErrors: false,
```
jwr1: I think showing errors should be on by default, cause this still affects things like if a login fails, if a button press or reply submission fails. If it were off by default, then those actions would silently fail.

```suggestion
    showErrors: true,
```

jwr1 (changes requested):
Yes!!! I've been needing this, especially the in app logger!

PR #241, opened 2025-07-24
Rename timeline to combined in code.

olorin99:
Renames instances of 'timeline' in the code to 'combined'.
Think I got every one.

### DIFF
--- a/lib/src/controller/profile.dart
+++ b/lib/src/controller/profile.dart
@@ -480,8 +480,9 @@ class ProfileOptional with _$ProfileOptional {
 }
 
 Object? _parseFeedDefaultCombinedSort(Map json, String name) {
-  if (name == 'feedDefaultTimelineSort') name = 'feedDefaultCombinedSort';
-  return json[name];
+  final current = json[name];
+  if (current != null) return current;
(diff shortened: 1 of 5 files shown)


lib/src/controller/profile.dart
```diff
@@ -474,3 +478,21 @@ class ProfileOptional with _$ProfileOptional {
     return copyWith(autoSwitchAccount: null, filterLists: null);
   }
 }
+
+Object? _parseFeedDefaultCombinedSort(Map json, String name) {
+  if (name == 'feedDefaultTimelineSort') name = 'feedDefaultCombinedSort';
+  return json[name];
+}
```
jwr1: I set the `feedDefaultTimelineSort` value before this update and then ran the app with this update and it didn't seem like this migration worked, as there was no `feedDefaultCombinedSort` set in the profile. I wouldn't mind just scrapping this migration anyway though, since it won't cause an error (like `feedDefaultView` would).

olorin99: Just pushed a that should fix it. If it still has problems I'll scrap it.

PR #242, opened 2025-07-24
Allow swiping fullscreen images to exit view.

olorin99:
Swiping fullscreen image up or down greater than 1/3 of the view height exits the view. Otherwise the image will animate back to center of view.

### DIFF
--- a/lib/src/widgets/image.dart
+++ b/lib/src/widgets/image.dart
@@ -1,6 +1,5 @@
 import 'dart:io';
 import 'dart:math';
-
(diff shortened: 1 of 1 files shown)


jwr1 (changes requested):
This doesn't seem to work with zooming in/out (of the image) anymore? Is there a way to support both?

olorin99:
This should work now. Used https://pub.dev/packages/extended_image which also has some other nice to haves like simple editing which might use down the road.

jwr1 (approved):
Looks good, thanks!

PR #248, opened 2025-08-06
Widescreen adjustments.

### LINKED ISSUE #204: Broken views in landscape mode
### Describe the bug

Since the last update, I've been having issues with the app. I use interstellar in landscape mode on my android tablet. Subscriptions/magazines plus favorites continuosly appear in the thread view and I can't get rid of them. This means that now i have three columns while using the app: the menu, the subscripions/magazines plus favorites and the threads temselves This wasn't an issue in version 0.9.2. I prefer landscape vs potrait mode. The only way to get rid of the subscription/magazines plus favorites menu is by switching to potrait mode, which is my preferred view of using my tablet. In landscape mode viewing images in full mode is also not possible. 

### Steps to Reproduce

Just open the app in landscape mode.

### Version

0.9.3

Android

Mbin

### Additional context

The problem is presently on lemmy, piefed and mbin.

olorin99:
- Fix fullscreen image settings also used for widescreen layouts. 
- Make nav drawer expandable/collapsable in widescreen layout. 
- Fix image zoom on mouse scroll in feed screen.

Closes #204.

### DIFF
--- a/lib/src/widgets/content_item/content_item.dart
+++ b/lib/src/widgets/content_item/content_item.dart
@@ -366,7 +366,8 @@ class _ContentItemState extends State<ContentItem> {
     return LayoutBuilder(
       builder: (context, constrains) {
         final hasWideSize = constrains.maxWidth > 800;
-        final isRightImage = hasWideSize;
+        final isRightImage = hasWideSize &&
+            !widget.fullImageSize;
 
         final double rightImageSize = hasWideSize ? 128 : 64;
 
--- a/lib/src/controller/profile.dart
+++ b/lib/src/controller/profile.dart
@@ -207,7 +207,7 @@ class ProfileRequired with _$ProfileRequired {
     alwaysShowInstance: false,
     coverMediaMarkedSensitive: true,
     fullImageSizeThreads: false,
-    fullImageSizeMicroblogs: true,
+    fullImageSizeMicroblogs: false,
     feedDefaultView: FeedView.threads,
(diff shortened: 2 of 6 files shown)


lib/src/widgets/content_item/content_item.dart
```diff
@@ -366,7 +366,8 @@ class _ContentItemState extends State<ContentItem> {
     return LayoutBuilder(
       builder: (context, constrains) {
         final hasWideSize = constrains.maxWidth > 800;
-        final isRightImage = hasWideSize;
+        final isRightImage = hasWideSize &&
+            !widget.fullImageSize;
```
jwr1, line 370: Actually, I just realized this change would make it so that microblogs will default to use the full image size even on desktop. I think we should have this option though, so maybe we should just change the microblogs to not use full image size by default?

olorin99: Does it? It seems to work fine for me. When 'Full image size for microblogs' setting is off it uses a right image with a wide window and a top cropped image with a narrow window. When the setting is on is uses the full image for both.

jwr1: Yes, but my point was that "Full image size for microblogs" is on by default, so microblogs take up an abnormal amount of space on desktop with the default setup.

olorin99: Ah right of course.

jwr1 (approved):

PR #250, opened 2025-08-12
Piefed feeds API

olorin99:
This should be good to merge. The main thing is lacking API for getting feeds/topics by name. I've used a workaround for now and it should work fine but might be good to wait for the rest of the API to be released.

Adds feeds and topics to explore screen. Can't currently search for specific feed/topic but can scroll through the list.
When creating a feed PieFed users will be given an option to create either a "Client feed", "Server feed", or "Server topic". Client feed will open the usual screen for adding inputs while the other two will open the explore screen for the corresponding input type.

When on a mbin/lemmy account saved piefed feeds are greyed out and inaccessible.


lib/src/screens/explore/explore_screen.dart
jwr1, line 1: If you can't use the search for feeds and topics at the moment, then the search bar should be greyed out when the feeds or topics category is selected. Same for the sorting and filtering, since they didn't seem to do anything either.

lib/src/screens/explore/explore_screen_item.dart
jwr1, line 1: The subscription button doesn't seem to do anything for feeds and topics. Does PieFed not provide an endpoint yet for that? If that's true, then it should probably be greyed out.

Additionally, all of the topics subscription count is zero, does that mean PieFed actually doesn't provide a subscription count for topics? If so, then maybe we should completely hide the subscription button for topics.

lib/src/controller/feed.dart
jwr1, line 1: To make it more forward compatible, instead of using `name` to store the PieFed's server feed ID plus the server domain, maybe store the feed `machine_name` (such as `arts-craft@piefed.social`) and then have a separate optional property (like `serverId`) which stores the numeric ID (which we could easily remove later). The bonus to this would be that while displaying the feed input, the UI will show the feed's name instead of its ID, which will make it much easier to identify.

lib/src/screens/explore/explore_screen_item.dart
jwr1, line 1: When manually creating a client feed and then selecting server feeds or topics, the checkbox UI state doesn't seem to update.

lib/src/screens/explore/explore_screen.dart
jwr1, line 1: Since the search, sort, and filter don't work for feeds and topics, it would make sense to hide them completely from the selection screen from when you are adding a server feed or server topic to the app

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -112,17 +125,172 @@ class _FeedSettingsScreenState extends State<FeedSettingsScreen> {
...
 }
 
+void newFeed(BuildContext context) {
+  if (context.read<AppController>().serverSoftware != ServerSoftware.piefed) {
+    pushRoute(context, builder: (context) => const EditFeedScreen(feed: null));
+  } else {
+    ContextMenu(
+      items: [
+        ContextMenuItem(
+          title: '${l(context).client} ${l(context).feed}',
```
jwr1: This should be a separate localization item; many languages don't use the same order for adjectives.

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -112,17 +125,172 @@ class _FeedSettingsScreenState extends State<FeedSettingsScreen> {
...
+            await pushRoute(
+              context,
+              builder: (context) => const EditFeedScreen(feed: null),
+            );
+            if (!context.mounted) return;
+            Navigator.pop(context);
+          },
+        ),
+        ContextMenuItem(
+          title: '${l(context).server} ${l(context).feed}',
```
jwr1: Same as above:

> This should be a separate localization item; many languages don't use the same order for adjectives.

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -112,17 +125,172 @@ class _FeedSettingsScreenState extends State<FeedSettingsScreen> {
...
+                if (!context.mounted) return;
+                context.read<AppController>().setFeed(title, feed);
+                Navigator.pop(context);
+                Navigator.pop(context);
+              },
+            ),
+          ),
+        ),
+        ContextMenuItem(
+          title: '${l(context).server} ${l(context).topic}',
```
jwr1: Same as above:

> This should be a separate localization item; many languages don't use the same order for adjectives.

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -112,17 +125,172 @@ class _FeedSettingsScreenState extends State<FeedSettingsScreen> {
...
-            ),
+            onTap: () => newFeed(context),
           ),
         ],
       ),
     );
   }
 }
 
+void newFeed(BuildContext context) {
```
jwr1, line 137: Would you be able to add a brief subtitle to each of the three types here that makes it clear the differences? I just think some people might be confused at first glance.

lib/src/screens/settings/feed_settings_screen.dart
```diff
@@ -95,7 +108,7 @@ class _FeedSettingsScreenState extends State<FeedSettingsScreen> {
                       await Navigator.of(context).push(
                         MaterialPageRoute(
                           builder: (context) => CreateScreen(
-                            initTitle: '[Feed] $name',
+                            initTitle: '[Feed] $entry.key',
```
jwr1: This change seems to have caused an issue. If I go to share a feed, it has the following post title:

```
[Feed] MapEntry(Rust: Feed(name: Rust, inputs: {FeedInput(name: privacy@programming.dev, sourceType: FeedSource.community), FeedInput(name: rust@programming.dev, sourceType: FeedSource.community), FeedInput(name: 48@piefed.social, sourceType: FeedSource.feed)})).key
```

lib/src/screens/settings/feed_settings_screen.dart
jwr1, line 1: I just realized you can't share an Interstellar feed from the markdown editor. Can you add support for that too?

lib/src/screens/explore/explore_screen_item.dart
jwr1, line 1: To make feed and topic items consistent with community and people items, can you make `title` use the feed's "display name" and `subtitle` use the feed's "machine name".

lib/src/controller/feed.dart
```diff
@@ -29,4 +30,14 @@ class Feed with _$Feed {
...
+
+  bool get clientFeed {
+    return !serverFeed;
+  }
+
+  bool get serverFeed {
+    return inputs.length == 1 &&
+        (inputs.firstOrNull?.sourceType == FeedSource.feed ||
+            inputs.firstOrNull?.sourceType == FeedSource.topic);
+  }
```
jwr1: Quick question about this. What makes the distinction between a feed with one piefed feed input vs multiple? As I was testing things out, I added a "server feed", then I went to edit it, then added an additional input, then I backed out to the feeds screen, and suddenly it wasn't considered a "server feed" anymore but instead a "client feed".

Instead of making a distinction here, should we just consider all feeds (in the context of Interstellar's client feeds), no matter the inputs, as client feeds? Just thinking out loud here.

olorin99: The distinction is that with only one input there isn't any merging/sorting client side. Alongside if the input is a piefed feed then it would show up as empty on mbin/lemmy so I thought it would be good to flag as different so mbin/lemmy can ignore it. Didn't really think of having multiple piefed feeds together, I was just thinking maybe having a piefed feed + some extra inputs, so maybe I should make that a check across all inputs in the feed rather than just the first. Something like
```
bool get serverFeed {
    return inputs.every((input) => 
        input.sourceType == FeedSource.feed || 
        input.sourceType == FeedSource.topic
    );
}
```
So if all the inputs are piefed feeds then mbin/lemmy can skip over the feed but if there are other supported inputs mixed then mbin/lemmy can at least show those.

olorin99:
I think I saw in the PieFed General matrix chat that rimu is planning on releasing 1.2 this friday though not confirmed. It's likely most instances will update quickly so it is probably safe to target 1.2. There are a couple other fallbacks for the API differences I had which could be removed as well.

lib/src/controller/feed.dart
```diff
@@ -13,6 +13,7 @@ class FeedInput with _$FeedInput {
   const factory FeedInput({
     required String name,
```
jwr1, line 14: I just realized this property is redundant right here, since the name is already contained in the database's key, similarly to profiles, accounts, servers, and filter lists. Can it be removed?

I believe this will be the final thing, and then we can merge.

jwr1 (approved):
Feel free to squash and merge whenever you're ready

PR #253, opened 2025-08-15
Reply inline option

olorin99:
Add setting to control whether reply controls are shown inline or on another screen.


jwr1:
It seems better, but the scrolling is super slow now on the drafts tab, and I get the following error:

```
Another exception was thrown: 'package:flutter/src/rendering/proxy_box.dart': Failed assertion: line 3047
pos 12: '!debugNeedsLayout': is not true.
```

PR #255, opened 2025-08-18
Add simple cache for feed inputs.

olorin99:
Switching between large feeds is likely to get rate limit errors due to fetching input ids each time. This should alleviate that a little.

### DIFF
--- a/lib/src/controller/controller.dart
+++ b/lib/src/controller/controller.dart
@@ -386,11 +386,6 @@ class AppController with ChangeNotifier {
       finder: Finder(filter: Filter.equals('account', key)),
     );
 
-    _feedCacheStore.delete(
-      db,
-      finder: Finder(filter: Filter.equals('account', key)),
-    );
-
     _rebuildProfile();
 
     _accounts.remove(key);
@@ -404,6 +399,11 @@ class AppController with ChangeNotifier {
           orElse: () => '',
         )
         .isEmpty) {
+      _feedCacheStore.delete(
+        db,
+        finder: Finder(filter: Filter.equals('server', keyAccountServer)),
+      );
+
       _servers.remove(keyAccountServer);
 
       await _serverStore.record(keyAccountServer).delete(db);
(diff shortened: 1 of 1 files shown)


lib/src/controller/controller.dart
```diff
@@ -714,6 +721,38 @@ class AppController with ChangeNotifier {
         null;
   }
 
+  Finder _feedCacheStoreFinder(String name) => Finder(
+    filter: Filter.and([
+      Filter.equals('account', _selectedAccount),
```
jwr1: It seems like using the server here instead would make more sense since the input IDs (communities and users) change per-server, not per-account.

lib/src/controller/controller.dart
```diff
@@ -714,6 +721,38 @@ class AppController with ChangeNotifier {
...
+      Filter.equals('name', name),
```
jwr1, line 727: Does `name` distinguish between the different input types? For example, would there be an issue if a community and a user happen to have the same name, or does it use the `!` and `@` prefixes like stars do?

lib/src/controller/controller.dart
```diff
@@ -714,6 +721,38 @@ class AppController with ChangeNotifier {
...
+    try {
+      final newValue = switch (source) {
+        FeedSource.community => (await api.community.getByName(name)).id,
+        FeedSource.user => (await api.users.getByName(name)).id,
+        _ => null,
+      };
+
+      if (newValue != null) {
+        await _feedCacheStore.add(db, {
+          'account': _selectedAccount,
```
jwr1: Same thing as above about using the server instead of the account.

lib/src/controller/controller.dart
```diff
@@ -384,6 +386,11 @@ class AppController with ChangeNotifier {
       finder: Finder(filter: Filter.equals('account', key)),
     );
 
+    _feedCacheStore.delete(
+      db,
+      finder: Finder(filter: Filter.equals('account', key)),
+    );
```
jwr1: Same thing as above about using the server instead of the account. Would want to move this down to the if statement where it's checked if the last account was removed from a server.

lib/src/controller/controller.dart
olorin99: It doesn't but I thought the name had to be unique across the fediverse. Can you have a user `name@instance.tld` and a community `name@instance.tld`?

jwr1: Yes, you can. Even on kbin.earth we have both a user and a magazine called `kbinEarth`.

olorin99: Ah TIL. I'll make the change.

PR #257, opened 2025-08-28
Adds default downloads directory option.

olorin99:
Adds default downloads directory option.
Adds longpress callback to loading icon button.

If default directory is set, tapping image download button downloads to default location else asks for location.
Long pressing the image download button will always ask for location.


lib/src/screens/settings/behavior_screen.dart
jwr1, line 1: It seems the ui state isn't being updated when you set (or reset) the default download directory. I have to exit out and reopen the behavior settings in order to see it updated.

PR #264, opened 2025-09-17
Cross posts

olorin99:
Gets cross post info for post.
Adds a button between post body and comments which opens a menu to allow navigating to cross posts.
Adds an option to allow showing comment sections from cross posts together similar to piefed web.
Adds an option to mark all cross posts as read when opening one of them.

Lemmy and PieFed don't return cross post info in the `/post/list` endpoint, only from the `/post` endpoint. This PR changes `PostPage` so it fetches the full post when opening the widget. This is skipped for Mbin since its unneeded.

### DIFF
--- a/lib/src/controller/profile.dart
+++ b/lib/src/controller/profile.dart
@@ -203,7 +203,7 @@ class ProfileRequired with _$ProfileRequired {
     markMicroblogsReadOnScroll: false,
     animationSpeed: 1.0,
     inlineReplies: true,
-    showCrosspostComments: false,
+    showCrosspostComments: true,
     markCrosspostsAsRead: false,
     appLanguage: '',
     themeMode: ThemeMode.system,


jwr1 (approved):
Nice, looks great! I especially like being able to see the cross-posted comments on the same screen. How would you feel about having that on by default? Should we make a poll on Matrix?

I do feel like the displayed information about the crossposts could be further improved, such as displaying score and comment count (and maybe showing voting buttons), but this is already great as is, and those could be worked on in a future release.

I do have one question, though: how does the infinite scrolling work when crosspost comments are displayed? I'm assuming as you scroll it loads all of the comments from the first post, and then once there are no more comments, it starts loading all of the comments from the second post, and so on?

olorin99:
Yeah showing showing the comments should probably be default. I don't imagine many people would turn it off.

I was planning on adding a context menu for each cross post but it started turning into a bit of a refactor so I decided to leave it for later. Ideally would use the `ContentItem` menu which has voting, report, bookmark, etc.

Each cross post section loads the initial request and then loads subsequent requests when scrolled into view. When the page is opened the comment sections don't take up much space vertically so they are all rendered to screen and start requesting comments but once the comments load they are pushed down off screen and so don't request more comments until scrolled back into view.

jwr1 (approved):

PR #274, opened 2025-09-21
Fix paging refresh.

olorin99:
Fixes paging controller refresh not refreshing to correct state.

Was causing explore to not reload properly when changing search input.


jwr1 (approved):
Ah thank you, I'll do a patch release in the next few days hoping we can find a few more bugs.

PR #276, opened 2025-09-21
AppImage build: simplify packages install, don't install ffmpeg

Samueru-sama:
This script automates the installing of the packages. 

You might wanna add `|| true` if you don't want the CI to fail when it is not possible to install the packages, in the end they are optional just to reduce size and not required to build the appimage.

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -153,30 +153,11 @@ jobs:
         if: false
         #if: matrix.target_os == 'linux'
         run: |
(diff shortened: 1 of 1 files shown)


Samueru-sama:
I updated the debloated packages repo and added checks to stop this from happening in the future: https://github.com/pkgforge-dev/archlinux-pkgs-debloated/pull/30

Of course right now the check does not work because the ffmpeg package is already in version 8, but once archlinux updates the package to version 8 this nonsense should happen ever again.

PR #292, opened 2025-09-26
Fix direct message page error for piefed.

### LINKED ISSUE #287: Direct messaging on PieFed is broken
### Describe the bug

Investigate why starting a new chat with someone on PieFed and Lemmy causes a grey screen with an error to appear.

I believe PieFed has worked on their DM APIs recently, so we'll definitely want to take a look at what's changed.

### Steps to Reproduce

_No response_

### Version

0.10.1

Android

PieFed

### Additional context

Requested in this comment: kbin.earth/m/interstellar/t/1925135/-/comment/8990665

olorin99:
Looks like this error was due to no messages with the user in the first returned page of messages which resulted in an empty participant list.

Closes #287

### DIFF
--- a/lib/src/screens/account/messages/message_thread_screen.dart
+++ b/lib/src/screens/account/messages/message_thread_screen.dart
@@ -95,10 +95,12 @@ class _MessageThreadScreenState extends State<MessageThreadScreen> {
     final myUsername = context.watch<AppController>().localName;
 
     final messageUser =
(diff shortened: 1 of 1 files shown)


jwr1 (approved):
Thank you, looks like it works! It seems like piefed DMs aren't reaching kbin.earth, but that's a PieFed or Mbin problem :)

PR #293, opened 2025-09-30
Context menu changes

olorin99:
- Adds context menu to cross post comment titles.
- Adds user/community submenus under content menu.
- Add bookmark menu item for piefed/lemmy.
- Add open in browser to user menu.

- Switch to using detailed user model in post model rather than the small user model.


jwr1:
> Switch to using detailed user model in post model rather than the small user model.

Does that actually work though? I just checked the Mbin API docs and it still says the `user` returned with a entry or microblog is `UserSmallResponseDto` (which is just the normal `UserModel` for Interstellar).

PR #294, opened 2025-10-01
Fix some nsfw blurring.

olorin99:
- Fix sensitive content setting not affecting embedded images.
- Add blur to nsfw videos, both post and embedded.


jwr1 (dismissed):
Looks good!

Just to confirm, while a video is playing, it won't be blurred will it? Just before it starts playing?

olorin99:
Good point, forgot to test that.

PR #298, opened 2025-10-21
Fix piefed mod menu not available

### LINKED ISSUE #297: Moderation actions not visible on PieFed
### Describe the bug

Moderation actions (pin, mark NSFW, delete, and ban) seem to no longer be available on PieFed through the overflow menu. I went to both the "Mod" feed view and also a magazine that I moderate, with no luck. I did open a post under the "Mod" view and one of the comments had the "Moderate" menu, but weirdly also had the Pin item in there (which I believe is only for posts).

### Steps to Reproduce

_No response_

### Version

0.10.1

Linux

PieFed

### Additional context

_No response_

olorin99:
Looks like rimu fixed the missing field on piefeds side. This fixes it on our side.
Also limits the items in the mod context menu for comments, so pin and mark as nsfw are no longer available from the comment mod menu.

Tested on crust.piefed.social.
Closes #297

### DIFF
--- a/lib/src/widgets/menus/content_menu.dart
+++ b/lib/src/widgets/menus/content_menu.dart
@@ -314,38 +314,42 @@ Future<void> showContentMenu(
           title: l(context).moderate,
           // onTap: () async => showModerateMenu(context, widget),
           subItems: [
-            ContextMenuItem(
-              title: l(context).pin,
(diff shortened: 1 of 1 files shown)


jwr1 (approved):
LGTM, thanks!

PR #304, opened 2025-11-08
Post layout update

### LINKED ISSUE #112: post appearance options
### Describe the feature

separate options to display poster username and magazine name above above post

### Additional context

_No response_

### LINKED ISSUE #208: Additional post views
### Describe the feature

Interstellar currently offers the default cards mode and the compact mode.
What I like from other threadiverse apps are their additional post views.
Especially Boost's "Cards 2.0" post view, which displays posts differently depending on whether they are link posts, picture posts, video posts or text posts:

Image

It is mainly this differentiation that I am missing from the available views on Interstellar - seeing link preview pictures stretch big like proper images looks weird for me, and seeing proper images compact seems like a nuisance to me.

### Additional context

_No response_

### LINKED ISSUE #214: Link images small
### Describe the bug

How it appears in interstellar
Image

How it appears in a lemmy client
Image

I assume there there’s a core difference here like displaying at native image resolution vs resizing to fill the feed width so maybe this is better listed as an enhancement vs bug. 

### Steps to Reproduce

Scroll through feed and look for any link posts. 

**Note** does not affect image posts (eg lemmy pictrs). 

### Version

0.9.3

iOS

Lemmy

### Additional context

_No response_

### LINKED ISSUE #269: full width view
### Describe the feature

Add a full width view for posts instead of the card format, horizontal space is already tight on mobile.

### Additional context

_No response_

olorin99:
Adds setting to present posts in the feed view as cards or not. Closes #269 

Separates some of the `ContentItem` out into smaller widgets and adds setting to allow changing the order of these widgets. Closes #208, closes #112 
The reorderable widgets are as follows
- Image
- Title
- Link
- Info (community, user, is nsfw, etc)
- Body

Also changed so when a post is a link and the image is less than 800px in width the image is treated as a thumbnail and drawn on the side of the post. Closes #214. 
Not sure if 800px is the best cuttoff point for this. I chose it based on the inverse of how currently if the page width is greater than 800 it moves all images over to the right. Based on some testing 800 does seem to work fine. Smaller link images are shown on the side while the large high def images are shown in full.

### DIFF
--- a/lib/src/screens/settings/display_screen.dart
+++ b/lib/src/screens/settings/display_screen.dart
@@ -128,17 +128,19 @@ class DisplaySettingsScreen extends StatelessWidget {
               children: [
                 Text(
                   '${l(context).settings_dividerThickness} : '
-                      '${ac.profile.dividerThickness.toStringAsFixed(1)}',
+                  '${ac.profile.dividerThickness.toStringAsFixed(1)}',
                 ),
                 Slider(
                   value: ac.profile.dividerThickness,
                   max: 10,
                   min: 0,
-                  onChanged: (newValue) => ac.updateProfile(
-                    ac.selectedProfileValue.copyWith(
-                      dividerThickness: newValue,
-                    ),
-                  ),
+                  onChanged: ac.profile.showPostsCards && !ac.profile.compactMode
+                      ? null
+                      : (newValue) => ac.updateProfile(
+                          ac.selectedProfileValue.copyWith(
+                            dividerThickness: newValue,
+                          ),
+                        ),
                   divisions: 50,
                 ),
               ],
--- a/lib/src/widgets/content_item/content_item.dart
+++ b/lib/src/widgets/content_item/content_item.dart
@@ -449,9 +449,7 @@ class _ContentItemState extends State<ContentItem> {
             }
 
             return Padding(
-              padding: widget.contentTypeName == l(context).comment
-                  ? const EdgeInsets.fromLTRB(12, 0, 12, 8)
-                  : const EdgeInsets.fromLTRB(12, 8, 12, 8),
+              padding: const EdgeInsets.fromLTRB(12, 8, 12, 8),
               child: Column(
                 mainAxisSize: MainAxisSize.min,
                 crossAxisAlignment: CrossAxisAlignment.stretch,
--- a/lib/src/screens/feed/post_comment.dart
+++ b/lib/src/screens/feed/post_comment.dart
@@ -368,6 +368,8 @@ class _PostCommentState extends State<PostComment> {
     );
 
     final menuWidget = IconButton(
+      padding: EdgeInsets.zero,
+      constraints: BoxConstraints(),
       icon: const Icon(Symbols.more_vert_rounded),
       onPressed: () {
         showContentMenu(
@@ -402,7 +404,7 @@ class _PostCommentState extends State<PostComment> {
             },
           ),
           child: Container(
-            padding: const EdgeInsets.fromLTRB(12, 0, 12, 0),
+            padding: const EdgeInsets.fromLTRB(12, 8, 12, 8),
             child: Row(
               mainAxisAlignment: MainAxisAlignment.spaceBetween,
               children: [


lib/src/widgets/image.dart
jwr1, line 1: I'm not sure if this was introduced in this pr or was it just already an issue (I think it was), but the hero animation is kind of janky when entering or exiting the full image view. For instance, a certain-sized image will become super large at the start of the transition animation.

lib/src/widgets/content_item/content_item.dart
jwr1, line 1: It seems like the `hero` id needs to be adjusted. I was continuously getting this error during testing:

```
Another exception was thrown: There are multiple heroes that share the same tag within a subtree.
```

jwr1, line 1: I think it would be beneficial to have a divider displayed between posts when the card mode is disabled, so it's easy to distinguish where one post ends and the next one begins. The compact mode already has such a divider too.

lib/src/screens/settings/display_screen.dart
jwr1, line 1: I'm assuming the post as card toggle has no effect while in compact mode? If that's the case, then the (card) setting should be disabled so as not to confuse people.

lib/src/screens/settings/post_layout.dart
```diff
@@ -0,0 +1,90 @@
...
+            onPressed: () => setState(() {
+              _postComponentOrder = ProfileRequired
+                  .defaultProfile
+                  .postComponentOrder
+                  .toList();
+              ac.updateProfile(
+                ac.selectedProfileValue.copyWith(
+                  postComponentOrder: _postComponentOrder,
+                ),
+              );
```
jwr1, line 50: Similar to your other pr, would it work to have the reset button set the value in the database to null?

lib/src/widgets/content_item/content_item.dart
jwr1, line 1: The "Post layout order" setting implies it will only change the layout of posts. The enum being called `PostComponent` also implies the same. However, I've noticed it also changes the layout of comments as well. Maybe comments should just ignore that setting and use the default order (Info, then Body)?

lib/src/widgets/content_item/content_item.dart
```diff
@@ -27,6 +25,8 @@ import 'package:provider/provider.dart';
 import 'package:interstellar/src/widgets/content_item/content_item_link_panel.dart';
 import 'package:simplytranslate/simplytranslate.dart';
 
+enum PostComponent { title, image, info, body, link }
```
jwr1, line 30: Do you think we should allow moving the action buttons around too, or would it be best to leave that out?

lib/src/widgets/content_item/content_item.dart
jwr1, line 1: The padding between components feels a bit squished or inconsistent. For instance I set Info to be at the bottom (on top of the actions buttons), and it feels like the user avatar and name is right on top of the comment icon (and other buttons). With Link positioned between Title and Body, the link top padding is noticeably smaller than the bottom padding.

lib/src/widgets/image.dart
olorin99: Yeah flutter can't animate between different image fits. I've changed it so the same fit is used for both sides of the animation.

lib/src/widgets/content_item/content_item.dart
olorin99: I did test this but it looked really weird with the action buttons in the middle of the layout so just left them at the bottom. It would be a simple change to allow moving them around too though so maybe we add it just in case.

lib/src/widgets/content_item/content_item.dart
olorin99: I also added a setting to control the thickness of the divider.

lib/src/widgets/content_item/content_item.dart
jwr1: No, it's alright. Especially if you already tried it and it looked weird.

lib/src/screens/settings/post_layout.dart
```diff
@@ -0,0 +1,90 @@
...
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(l(context).settings_postLayoutOrder),
+        actions: [
+          IconButton(
+            onPressed: () => setState(() {
+              _postComponentOrder = ProfileRequired
+                  .defaultProfile
+                  .postComponentOrder
+                  .toList();
```
jwr1, line 45:
```suggestion
```

olorin99: Like with the other PR this is used as state for the reorderable widget.

lib/src/screens/settings/display_screen.dart
```diff
@@ -105,6 +122,28 @@ class DisplaySettingsScreen extends StatelessWidget {
...
+                  onChanged: (newValue) => ac.updateProfile(
+                    ac.selectedProfileValue.copyWith(
+                      dividerThickness: newValue,
+                    ),
+                  ),
+                  divisions: 50,
+                ),
+              ],
+            ),
+          ),
```
jwr1, line 148: Looks great! I just have one last thing, can you disable this setting when it's not in use? I guess that would mean having it disabled while compact mode is disabled and while card mode is enabled.

jwr1 (approved):
Thank you!

Issue #84, opened 2025-01-17
Recurring Oauth authorization error


haverholm:
### Describe the bug

Having used Interstellar for a couple of months now, I've noticed that every 7-10 days or so, the app returns an Oauth error from my mbin account on kbin.earth:

`Oauth authorization error (invalid_request): The refresh token is invalid..`

I suspect it's either an mbin or instance specific issue as this doesn't occur with lemmy accounts I have set up in the app.

My solution has been deleting the mbin account and reauthorizing, but that gets tired in the long run.

### Version

0.7.1 (F-droid)

### What platform(s) are you seeing the problem on?

Android

jwr1:
Thanks for creating an issue. I personally have only run into this issue a few rare times, and that not being anytime recent. I will definitely investigate the problem though.

You are correct about it being an Mbin only issue; since this is an oauth problem, then Lemmy wouldn't have the same issue since it doesn't use oauth (it uses jwt instead).

jwr1:
I looked into it briefly, and from what I can tell, everything is fine on Interstellar's end. I manually triggered credential refresh a few times, and it worked good. I'm thinking maybe Mbin's refresh token expiration time is shorter than it should be.

haverholm:
I've tried to determine if there is any pattern to the error messages, and apparently there isn't. That indicates to me that it's not a token expiring at regular intervals? 🤔

qupada:
I've run into this a few times with my account on fedia.io, though at a much lower rate than described in the original report.  I bought a new phone around 60 days ago and had an expiry for the first time today.

I'd personally be happy enough if I just got bumped to a login screen in these situations, but as has been noted it's moderately difficult to force it to re-authenticate you.

joshg253:
Just hit this for the first time with my Fedia.io account. It's been working fine for months. My two PieFed alts are still working fine in the app. Fedia.io working fine via web.

PR #308, opened 2025-11-13
Allow user to quickly relogin on oauth error.

olorin99:
Moves login callback to controller.
Add error page with button to allow user to quickly relogin on oauth error.
Partially addresses #84 but we should still try and figure out the underlying issue.


jwr1 (approved):
Thank you, it worked great for me! This is definitely a needed feature!

PR #311, opened 2025-11-17
Comment images

### LINKED ISSUE #261: image picker for comments
### Describe the feature

currently you can only add media through md embeds in comments, it would be nice if I could add images directly like I can in posts.

### Additional context

_No response_

olorin99:
Markdown image action can now open system image picker and uploads image.
Available image upload locations
- platform (lemmy/piefed)
- catbox.moe
- imglink.io  

Default upload location can be chosen in behaviour settings.

Add image selector in comment reply screen when on mbin. This will upload the image to mbin while the markdown editor option will default to catbox.moe.

Closes #261

### DIFF
--- a/lib/src/api/images.dart
+++ b/lib/src/api/images.dart
@@ -7,7 +7,7 @@ import 'package:path/path.dart';
 import 'package:http_parser/http_parser.dart';
 import 'package:image_picker/image_picker.dart';
 
-enum ImageStore { platform, catbox }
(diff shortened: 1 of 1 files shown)


jwr1 (approved):
Looks good, thanks!

It seems like the markdown actions have some issues (like selector values going out of range), but that was already present and a completely separate issue; maybe I could look at it later.

PR #313, opened 2025-11-24
Misc fixes

### LINKED ISSUE #263: Cannot switch accounts
### Describe the bug

After logging in, Switch accounts opens a blank grey box on the bottom half of the screen with the message "Instance of 'DiagnosticsProperty<void>" after clicking outside the box.

Image
Image

### Steps to Reproduce

1. Open settings
2. Select Switch accounts

### Version

v0.9.3

Android

PieFed

### Additional context

_No response_

olorin99:
Fix create account with null handle.
Fix null access when empty account on account selection screen. Might fix #263 but not sure if same bug.

### DIFF
--- a/lib/src/screens/settings/account_selection.dart
+++ b/lib/src/screens/settings/account_selection.dart
@@ -89,8 +89,9 @@ class _AccountSelectWidgetState extends State<AccountSelectWidget> {
               ...ac.accounts.keys
                   .where(
                     (account) =>
-                        !widget.onlyNonGuestAccounts ||
-                        account.split('@').first.isNotEmpty,
+                        (!widget.onlyNonGuestAccounts ||
+                            account.split('@').first.isNotEmpty) &&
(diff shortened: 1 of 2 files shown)


lib/src/screens/settings/account_selection.dart
```diff
@@ -89,8 +89,9 @@ class _AccountSelectWidgetState extends State<AccountSelectWidget> {
               ...ac.accounts.keys
                   .where(
                     (account) =>
-                        !widget.onlyNonGuestAccounts ||
-                        account.split('@').first.isNotEmpty,
+                        (!widget.onlyNonGuestAccounts ||
+                            account.split('@').first.isNotEmpty) &&
+                        account.isNotEmpty,
```
jwr1, line 94: Wait a second, I took another look at this because I was running into this issue, but there should never be an account that's just an empty handle anyway. So we shouldn't even have to check it here.

PR #314, opened 2025-11-27
Tags flairs

olorin99:
Adds support for Piefed post flairs and user flairs/tags.

Adds user tags.
- Multiple tags can be assigned to each user.
- Tags can be created and destroyed from tags screen in settings
- Tags can be added or removed from a user via the user menu.


lib/src/controller/database.dart
```diff
@@ -371,6 +372,33 @@ class Drafts extends Table {
...
+  TextColumn get tag => text()();
+  IntColumn get backgroundColor => integer().map(const ColorConverter()).clientDefault(() => Color.from(alpha: 1, red: 1, green: 1, blue: 1).value32bit)();
+  IntColumn get textColor => integer().map(const ColorConverter()).clientDefault(() => Color.from(alpha: 1, red: 0, green: 0, blue: 0).value32bit)();
+}
+
+class UserTags extends Table {
+  IntColumn get id => integer().autoIncrement()();
+  TextColumn get user => text()();
+  IntColumn get tagId => integer().references(Tags, #id, onDelete: KeyAction.cascade)();
+}
```
jwr1, line 401: Assuming you're only supposed to have 1 of the same tag per user, then we shouldn't need `id` here and can make `user` and `tagId` the composite primary key, right?

lib/src/controller/database.dart
```diff
@@ -371,6 +372,33 @@ class Drafts extends Table {
...
+
+  @override
+  int toSql(Color color) {
+    return color.value32bit;
+  }
+}
+
+class Tags extends Table {
+  IntColumn get id => integer().autoIncrement()();
+  TextColumn get tag => text()();
```
jwr1: Should this have a unique constraint on it?

lib/src/screens/feed/post_comment.dart
jwr1, line 1: Would there be a better way to display user tags without having to wrap every instance of a post or comment to add in the tags? For example, if we instead moved this to the ContentItem widget, and checked if the passed in username matched with any local tags, then it would add in the extra tags next to the PieFed flairs.

The way it's done right now just seems like it would be easy to mess up in the future (like forgetting to use `applyUserTagsComment` where it should be).

Another issue I see with it is that post flairs and local user tags are merged together and treated as the same. What happens if we allow editing the flairs of a post later on, and the local tags show up just the same as the flairs and are indistinguishable?

olorin99: Doing it directly in content item is indeed much easier and cleaner. There is a little pop in when the tags are fetched but its not easily noticeable so no worries.

The post flairs and user tags are stored separately. I've started working on editing post flairs and user flairs getting in the way shouldn't be an issue. It would be nice to have a visual difference between them though. Maybe a border around one of them with the same colour as the text would work to differentiate them.

jwr1 (dismissed):
I'll go ahead and merge this, but I do have a few changes I want to personally make. I'll do it in a separate pr though so you can review them.

lib/src/screens/feed/post_page.dart
```diff
@@ -61,14 +61,23 @@ class _PostPageState extends State<PostPage> {
...
+          }.then(
+            (post) async => post.copyWith(
+              user: post.user.copyWith(
+                tags: [
+                  ...post.user.tags,
+                  ...(await ac.getUserTags(post.user.name)),
+                ],
+              ),
+            ),
+          );
```
jwr1: Is this still needed?

PR #315, opened 2025-11-29
Add ability to create thread and microblog cross posts

jwr1:
There's a new menu item for posts "Cross post" that will help you create a cross post of an existing post. It will copy over the post title and link (or image src). For the body, it will be "Cross posted from {APID}", and then the original body beneath wrapped in a markdown quote.

The most prominent question I had while working on this was whether the correct terminology was "crosspost", "cross post", or "cross-post" :). I ultimately went with "cross post" and updated almost everything else to match. I even changed the database columns (without migration), so you'll need to recreate your database for this.


olorin99 (dismissed):
Would it make sense to put "Share" and "Cross Post" together in a submenu since they're somewhat similar, or do you think they're different enough to be separate? Otherwise looks good.

olorin99:
Yeah that sounds good. Could also add options to share either the local or remote version of the post if it originates from another server.

PR #317, opened 2025-12-02
Add support for piefed polls.

olorin99:
Adds support for piefed polls.
- Show poll posts with button to toggle whether results are shown.
- Vote on posts. Seems like it is a limitation on piefed that you can't change a vote for single choice polls and you can't remove a vote for multichoice polls.
- Create polls. Can add options, set the poll mode and set the duration of the poll.

I don't believe the polls API has made it to stable yet so polls only work on crust.piefed.social for now.

### DIFF
--- a/lib/src/screens/feed/create_screen.dart
+++ b/lib/src/screens/feed/create_screen.dart
@@ -4,6 +4,7 @@ import 'package:image_picker/image_picker.dart';
 import 'package:interstellar/src/controller/controller.dart';
 import 'package:interstellar/src/controller/server.dart';
 import 'package:interstellar/src/models/community.dart';
+import 'package:interstellar/src/models/post.dart';
 import 'package:interstellar/src/screens/explore/community_owner_panel.dart';
 import 'package:interstellar/src/screens/explore/community_screen.dart';
 import 'package:interstellar/src/utils/language.dart';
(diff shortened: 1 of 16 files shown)


lib/src/widgets/content_item/poll.dart
```diff
@@ -0,0 +1,140 @@
...
+          ),
+          Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              Text('$totalVotes votes'),
+              ElevatedButton(
+                onPressed: () => setState(() {
+                  _showResults = !_showResults;
+                }),
+                child: Text('Show results'),
```
jwr1: This should be localized, and should be either "Show results" or "Hide results" based on the `_showResults` variable.

lib/src/widgets/content_item/poll.dart
```diff
@@ -0,0 +1,140 @@
...
+                                _choices =
+                                    post.poll?.choices.toList() ?? _choices;
+                              });
+                            },
+                          ),
+                        if (!widget.poll.multiple)
+                          Radio<int>(
+                            value: choice.id,
+                            enabled: _selectedAnswer == null,
+                          ),
```
jwr1: The radio and checkboxes should be disabled if you are not signed in, as they are currently clickable while browsing as guest

lib/src/screens/feed/create_screen.dart
```diff
@@ -194,6 +200,61 @@ class _CreateScreenState extends State<CreateScreen> {
...
+            final mode = await pollMode(
+              context,
+            ).askSelection(context, _pollMode);
+            if (mode == null) return;
+            setState(() {
+              _pollMode = mode;
+            });
+          },
+          trailing: Text(pollMode(context).getOption(_pollMode).title),
+        ),
```
jwr1: It seems like we could replace this section with a simple toggle switch, something like: "Allow multiple selecitons".

jwr1 (changes requested):
Looks pretty cool! Do you have any information on when the API will be moved to stable?

PR #320, opened 2025-12-04
Add links for local and remote urls to context menu

jwr1:
This will add source links to post, comment, community, and user context menus. Has both a local link and a remote link (only local if it's the source). This replaces any current "open in browser" and "share" menu items.

The link buttons function the same as current post links, so if you click on it, it will open in your browser, and if you long-press or right-click over it, it will give you more options. These include viewing/selecting the link, opening it in a browser, opening it in a WebView, copying it, or sharing it.

image

image


lib/src/models/user.dart
```diff
@@ -75,6 +76,7 @@ abstract class DetailedUserModel with _$DetailedUserModel {
           : NotificationControlStatus.fromJson(
               json['notificationStatus'] as String,
             ),
+      apId: json['apProfileId'] as String,
```
olorin99: Apparently apProfileId is nullable so should accept null values here. I got an error when I tried to login. Seems `/users/me` doesn't include this.

lib/src/models/user.dart
```diff
@@ -143,6 +148,7 @@ abstract class UserModel with _$UserModel {
     avatar: mbinGetOptionalImage(json['avatar'] as JsonMap?),
     createdAt: optionalDateTime(json['createdAt'] as String?),
     isBot: (json['isBot'] ?? false) as bool,
+    apId: json['apProfileId'] as String,
```
olorin99: Likewise here

lib/src/models/community.dart
```diff
@@ -93,6 +94,7 @@ abstract class DetailedCommunityModel with _$DetailedCommunityModel {
           : NotificationControlStatus.fromJson(
               json['notificationStatus'] as String,
             ),
+      apId: json['apProfileId'] as String,
```
olorin99: Also nullable

lib/src/widgets/context_menu.dart
```diff
@@ -85,9 +88,60 @@ class ContextMenu {
...
           ),
+          if (links.isNotEmpty)
+            Padding(
+              padding: const EdgeInsets.only(top: 8, left: 4, right: 4),
+              child: Row(
+                mainAxisAlignment: MainAxisAlignment.center,
+                children: links
+                    .map(
+                      (link) => Flexible(
+                        child: Card.outlined(
```
olorin99, line 101: Is there a way to make is obvious that these have secondary actions? Right now its not clear that you can long press to open a share menu. I can't really think of anything but it would be helpful.

olorin99, line 101: Maybe also add indicators to show what the links refer to. Something like `Local: ${link}`/`Remote: ${link}`. Or maybe with icons. A fediverse icon next to the remote link would help show what it is.

jwr1: I'm not sure, but the current post links don't have much indication that there are secondary actions either. I suppose I was hoping that since these links and the post links look pretty similar, people would guess they could use the secondary actions likewise.

If you have any suggestions to make it clear, I'm all for it.

jwr1: I think icons are a good idea :+1:.

jwr1:
How's this look? It shows a home icon for the local link, and fediverse icon for the remote link. If they are one and the same, then it shows both icons.

lib/src/widgets/context_menu.dart
olorin99: Yeah can't really think of anything. UI design is hard, can never really guess how people will interpret something.

PR #323, opened 2025-12-04
Set app user agent to "Interstellar/{APP_VERSION}"

jwr1:


lib/src/utils/http_client.dart
```diff
@@ -0,0 +1,42 @@
...
+class JwtHttpClient extends http.BaseClient {
+  final String _jwt;
+
+  final http.Client _httpClient;
+
+  JwtHttpClient(this._jwt, this._httpClient);
+
+  @override
+  Future<http.StreamedResponse> send(http.BaseRequest request) async {
+    request.headers['authorization'] = 'Bearer $_jwt';
```
melroy89, line 33: Maybe also add `request.headers['User-Agent']`?

jwr1: Thank you for checking, but that is covered. `JwtHttpClient` requires passing in an `httpClient`, and in `controller.dart` where `JwtHttpClient` is instantiated, `appHttpClient` is passed in (which is a `UserAgentHttpClient`).

melroy89: ah right now I see it, as a second parameter.. my bad. I was triggered by the fact it was also extending from the BaseClient, rather than extending from UserAgentHttpClient

olorin99 (dismissed):

PR #325, opened 2025-12-06
Allows adding and editing post flairs on piefed.

olorin99:
Can add post flairs when creating a post on piefed. Moderators or the post creator can also edit flairs via the context menu.

### DIFF
--- a/lib/src/models/post.dart
+++ b/lib/src/models/post.dart
@@ -346,7 +346,7 @@ abstract class PostModel with _$PostModel {
               .toList() ??
           [],
       flairs: (postView['flair_list'] as List<dynamic>?)?.map((flair) => Tag(
-        id: -1,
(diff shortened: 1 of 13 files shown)


lib/src/models/post.dart
```diff
@@ -346,7 +346,7 @@ abstract class PostModel with _$PostModel {
               .toList() ??
           [],
       flairs: (postView['flair_list'] as List<dynamic>?)?.map((flair) => Tag(
-        id: -1,
+        id: flair['id'] as int,
```
jwr1, line 349: Considering `Tag` is a database object, is it possible for `id` here to conflict with anything?

olorin99: Shouldn't do. The post flairs are stored separate from user tags and shouldn't go anywhere near the database.

PR #326, opened 2025-12-08
Some fixes/enhancements for polls

olorin99:
- Disable voting on poll after poll expires.
- Show time until poll expiration.
- During poll creation allow reordering choices.
- Change "Unlimited" poll duration option to have a duration of 100 years. Turns out passing a null expiration date defaults to 3 days. Figure 100 years is long enough to count as "unlimited".

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -569,14 +569,6 @@
       }
     }
   },
-  "pollExpiry_years": "{count} {count, plural, =1{year} other{years}} left",
-  "@pollExpiry_years": {
-    "placeholders": {
-      "count": {
-        "type": "int"
-      }
-    }
-  },
   "pollExpiry_days": "{count} {count, plural, =1{day} other{days}} left",
   "@pollExpiry_days": {
     "placeholders": {
--- a/lib/src/screens/feed/create_screen.dart
+++ b/lib/src/screens/feed/create_screen.dart
@@ -624,5 +624,5 @@ SelectionMenu<Duration?> pollDuration(BuildContext context) =>
       SelectionMenuItem(value: Duration(days: 1), title: l(context).pollDuration_days(1)),
       SelectionMenuItem(value: Duration(days: 3), title: l(context).pollDuration_days(3)),
       SelectionMenuItem(value: Duration(days: 7), title: l(context).pollDuration_days(7)),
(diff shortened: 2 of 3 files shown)


lib/l10n/app_en.arb
```diff
@@ -545,6 +569,46 @@
...
     }
   },
+  "pollExpiry_years": "{count} {count, plural, =1{year} other{years}} left",
+  "@pollExpiry_years": {
+    "placeholders": {
+      "count": {
+        "type": "int"
+      }
+    }
+  },
```
jwr1: I can't imagine many polls setting their duration to an amount in years, so I think this is unnecessary (especially since the max on PieFed's web UI is only 7 days). We could just cap the units at days (so 1 year would display as 365 days).

lib/src/screens/feed/create_screen.dart
```diff
@@ -599,12 +617,12 @@ class _CreateScreenState extends State<CreateScreen> {
...
-      SelectionMenuItem(value: Duration(days: 7), title: '7 days'),
-      SelectionMenuItem(value: Duration(), title: 'Unlimited'),
+      SelectionMenuItem(value: Duration(minutes: 30), title: l(context).pollDuration_minutes(30)),
+      SelectionMenuItem(value: Duration(hours: 1), title: l(context).pollDuration_hours(1)),
+      SelectionMenuItem(value: Duration(hours: 6), title: l(context).pollDuration_hours(6)),
+      SelectionMenuItem(value: Duration(hours: 12), title: l(context).pollDuration_hours(12)),
+      SelectionMenuItem(value: Duration(days: 1), title: l(context).pollDuration_days(1)),
+      SelectionMenuItem(value: Duration(days: 3), title: l(context).pollDuration_days(3)),
+      SelectionMenuItem(value: Duration(days: 7), title: l(context).pollDuration_days(7)),
+      SelectionMenuItem(value: Duration(days: 365 * 100), title: 'Unlimited'),
```
jwr1: It seems like 1 year (365 days) would be sufficient as the max value (and then update the title). Especially considering the max duration on PieFed's web UI is only 7 days.

```suggestion
      SelectionMenuItem(value: Duration(days: 365), title: l(context).pollDuration_days(365)),
```

jwr1 (approved):

PR #345, opened 2025-12-22
Modlog

olorin99:
Adds modlog for Mbin and lemmy (Piefed doesn't have API support yet).
Modlog is available from community menu for Mbin and from both community and user menus for Lemmy.
Also have a instance level modlog under the 'About' page in settings. We should probably add an instance page with instance info (admins, rules, etc) where this would be better placed but for now I figured under About made most sense.

Still needs a lot of testing but some feedback would be good. Should I include the deleted post/comment in the UI since you sort of expect to see unpleasant stuff in the modlog or should I obscure them.

Screenshot from 2025-12-22 23-00-46

### DIFF
--- a/lib/src/widgets/content_item/content_info.dart
+++ b/lib/src/widgets/content_item/content_info.dart
@@ -52,61 +52,46 @@ class ContentInfo extends StatelessWidget {
   Widget build(BuildContext context) {
     final warning = filterListWarnings == null || filterListWarnings!.isEmpty
         ? null
-        : Padding(
-            padding: const EdgeInsets.only(right: 10),
-            child: Tooltip(
-              message: l(
-                context,
-              ).filterListWarningX(filterListWarnings!.join(', ')),
-              triggerMode: TooltipTriggerMode.tap,
-              child: const Icon(
-                Symbols.warning_amber_rounded,
-                color: Colors.red,
-              ),
-            ),
+        : Tooltip(
+            message: l(
+              context,
+            ).filterListWarningX(filterListWarnings!.join(', ')),
+            triggerMode: TooltipTriggerMode.tap,
+            child: const Icon(Symbols.warning_amber_rounded, color: Colors.red),
           );
 
     final pinned = !isPinned
         ? null
-        : Padding(
-            padding: const EdgeInsets.only(right: 10),
-            child: Tooltip(
-              message: l(context).pinnedInCommunity,
-              triggerMode: TooltipTriggerMode.tap,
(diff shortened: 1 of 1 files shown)


jwr1:
Sure, I'm a bit busy this week, but I'll take a look when I get a chance! At first glance, since the mod log is quite a dense view in general, maybe we'd want to ditch the cards and just use dividers between items (or at least see what it looks like).

olorin99:
Removed the cards and added some dividers. I reckon this does look better.
Screenshot from 2025-12-27 23-54-15

jwr1:
Yeah, it does seem like we'd need to remove the dividers that are within each report, though. It took me a minute while looking at the screenshot, because it looked like the sections with "Content:" were a whole separate report. Besides that I think it looks pretty good, and I like the color coded flairs that are used for the different mod log types.

olorin99:
Think I've gotten everything. Mbin has less functionality as some things aren't returned in the API response. Mod added/removed and post pinned/unpinned won't link to the corresponding item as the user and post aren't available in the response. I think everything else should work though.

lib/src/api/moderation.dart
```diff
@@ -10,6 +11,90 @@ const _postTypeMbinComment = {
...
+  postDeleted,
+  postRestored,
+  commentDeleted,
+  commentRestored,
+  postPinned,
+  postUnpinned,
+  post_deleted,
+  post_restored,
+  post_comment_deleted,
+  post_comment_restored,
```
jwr1: These should be camel case right?

olorin99:
Made an issue [here](https://github.com/MbinOrg/mbin/issues/1931).

The user/community info should wrap now when there's not enough space. 
Screenshot from 2026-01-03 22-07-41

It's the same widget thats used for posts and comments so the user/community will wrap there too.
Screenshot from 2026-01-03 22-06-49

jwr1:
> Made an issue [here](https://github.com/MbinOrg/mbin/issues/1931).

Thank you!

> The user/community info should wrap now when there's not enough space.

I could be remembering wrong, but I believe previously the content info widget should have truncated the community and user names to prevent overflowing, so I thought we could just do the same here. I could work on that though, unless you want to.

olorin99:
That would be better. I looked through the diff of #304 with the post layout change and couldn't see where the user/community names would be truncated and can't think of any other PRs where it might have changed but if you have an idea of how to do it then that would be great.

olorin99:
Awesome thanks. Was just missing correct placement of `Expanded` and `Flexible`.

PR #358, opened 2026-01-22
Navigation update

olorin99:
I believe I've switched all the navigation across to using auto_route now. Next step is to make sure all the paths make sense and that path parameters match up.

I've also started on making sure it works on web. Two biggest issues so far seem to be with file io and checking the platform. Apparently `Platform.isAndroid`, `Platform.isLinux` etc don't work on web so I've had to add some util functions that check if it is web before checking the platform. Haven't switched out all the platform checks yet so some pages might throw errors but the main feed should show up on web now.

It looks like a really big diff but most (~13K) of the added lines come from adding a file for drift to work on web. 'web/drift_worker.dart.js'

### DIFF
--- a/lib/src/controller/controller.dart
+++ b/lib/src/controller/controller.dart
@@ -512,7 +512,9 @@ class AppController with ChangeNotifier {
         final result = Uri.parse(
           await FlutterWebAuth2.authenticate(
             url: authorizationUrl.toString(),
-            callbackUrlScheme: oauthRedirectUri.scheme,
+            callbackUrlScheme: PlatformIs.linux || PlatformIs.windows
+                ? oauthRedirectUri.toString()
+                : oauthRedirectUri.scheme,
           ),
         ).queryParameters;
 


lib/src/widgets/redirect_listen.dart
```diff
@@ -24,29 +27,37 @@ class RedirectListener extends StatefulWidget {
...
   Future<Uri> _listenForAuth() async {
-    HttpServer server = await HttpServer.bind(_redirectHost, _redirectPort);
-    await launchUrl(widget.initUri);
-    final req = await server.first;
+    if (!PlatformUtils.isWeb) {
+      _httpServer = await HttpServer.bind(_redirectHost, _redirectPort);
+      _httpServer?.listen((req) async {
+        req.response.statusCode = 200;
+        req.response.headers.set('content-type', 'text/plain');
+        req.response.writeln(l(context).redirectReceivedMessage);
```
jwr1: With the `flutter_web_auth_2` package, do we still need this here?

lib/src/api/oauth.dart
```diff
@@ -23,14 +24,18 @@ const oauthScopes = [
 Future<String> registerOauthApp(String instanceHost) async {
   const path = '/api/client';
 
+  final redirectUrl = PlatformUtils.isWeb
+      ? 'http://${Uri.base.host}:${Uri.base.port}/auth.html'
```
jwr1: Is this meant to be hard coded with `http`?

lib/src/controller/controller.dart
```diff
@@ -502,18 +507,18 @@ class AppController with ChangeNotifier {
           tokenEndpoint,
         );
 
+        final redirectUrl = PlatformUtils.isWeb
+            ? 'http://${Uri.base.host}:${Uri.base.port}/auth.html'
+            : redirectUri;
```
jwr1: This looks like the same logic as from `oauth.dart`. Could you combine it into one global variable? Maybe replacing the old `redirectUri` variable if it's not needed.

lib/src/screens/account/messages/message_thread_screen.dart
jwr1, line 1: What is the additional `userId` parameter here needed for? The message threads should be primarily represented by `threadId`.

lib/src/widgets/redirect_listen.dart
jwr1, line 1: Also, do we still need to manually create a web view, or will `flutter_web_auth_2` do that as well?

lib/main.dart
```diff
@@ -25,15 +25,15 @@ void main() async {
...
+  if (PlatformUtils.isDesktop) {
     await windowManager.ensureInitialized();
 
     // Get smallest dimensions of available displays and set minimum window
     // size to be a 16th of those dimensions.
     final screenSize = PlatformDispatcher.instance.displays
         .map((display) => display.size)
         .reduce((a, b) => Size(min(a.width, b.width), min(a.height, b.height)));
-    final minWindowSize = screenSize / 4;
+    final minWindowSize = screenSize / 8;//4;
```
jwr1:
```suggestion
    final minWindowSize = screenSize / 8;
```

We probably don't need this comment here.

lib/src/controller/router.dart
jwr1, line 1: It looks like the file that gets generated from this is not in `.gitignore`. We aren't supposed to check it in though right?

lib/src/utils/utils.dart
```diff
@@ -325,3 +327,53 @@ Future<void> pushRoute(
...
+  }
+
+  static bool get isLinux {
+    if (kIsWeb) {
+      return false;
+    } else {
+      return Platform.isLinux;
+    }
+  }
+}
```
jwr1, line 343:
```suggestion
class PlatformUtils {
  static const bool isWeb = kIsWeb;

  static final bool isAndroid = !isWeb && Platform.isAndroid;
  static final bool isIOS = !isWeb && Platform.isIOS;
  static final bool isLinux = !isWeb && Platform.isLinux;
  static final bool isMacOS = !isWeb && Platform.isMacOS;
  static final bool isWindows = !isWeb && Platform.isWindows;

  static final bool isMobile = isAndroid || isIOS;
  static final bool isDesktop = isLinux || isMacOS || isWindows;
}
```

Just to make it more concise, and maybe ever so slightly more performant.

lib/src/controller/router.dart
jwr1, line 1: This file is important enough we could move it outside the `utils` directory.

lib/src/controller/controller.dart
```diff
@@ -141,17 +144,19 @@ class AppController with ChangeNotifier {
       .update(database.miscCache)
       .write(MiscCacheCompanion(expandNavDomains: Value(value)));
 
-  Future<File> get logFile async {
+  Future<File?> get logFile async {
+    if (kIsWeb) return null;
```
jwr1:
```suggestion
    if (PlatformUtils.isWeb) return null;
```

Just for consistency.

lib/src/controller/database.dart
```diff
@@ -472,6 +484,7 @@ Future<void> deleteTables() async {
 }
 
 Future<bool> migrateDatabase() async {
+  if (kIsWeb) return false;
```
jwr1:
```suggestion
  if (PlatformUtils.isWeb) return false;
```

Just for consistency.

lib/src/controller/router.dart
jwr1, line 1: Should we try to make some of these routes more consistent with existing frontends? For example, almost every platform uses /u/namehere for users and /c/namehere for communities (except Mbin of course).

I also think there's an issue with the numeric id routes (threads, microblogs, comments), where they are not server agnostic, which would be an issue if you tried to copy-paste a link from one server to another. Some Lemmy frontends solve this by having the server domain in the path. I also think it would be nice to have the community name in the path as well.

Just as a reference, here are a few example links from other frontends:

Voyager:
`https://vger.app/posts/lemmy.zip/u/Stamets@lemmy.dbzer0.com`
`https://vger.app/posts/lemmy.zip/c/whitepeopletwitter@sh.itjust.works/comments/57940056`

Blorp:
`https://blorpblorp.xyz/home/c/technology@lemmy.world/posts/https%3A%2F%2Flemmy.world%2Fpost%2F42282750`
(blorp interestingly uses the AP-id in the URL)

Alexandrite:
`https://a.lemmy.world/lemmy.world/post/42382934`

I realize these would probably be some pretty big changes though, so we could definitely save it for a different PR, especially considering this one is already huge.

lib/src/screens/account/messages/message_thread_screen.dart
olorin99: The `userId` is used mainly for filtering messages on Lemmy and Piefed and for creating a new thread on Mbin. Although I believe Piefed added a endpoint to fetch a thread similar to Mbin so it might be worthwhile changing how messages work with Piefed.

lib/src/widgets/redirect_listen.dart
olorin99: I believe it's needed for desktop but not for web or mobile targets. I'll do some testing.

lib/src/widgets/redirect_listen.dart
olorin99: That should be handled for us too.

lib/src/controller/router.dart
olorin99: Yeah it would be good to be consistent with the other frontends. I think blorp's way of using the apid is probably the best from a compatibility with different servers standpoint. Especially for sharing links. But we would have to support fetching via the apid rather than server specific id, so maybe down the road.

Should we do something like
`https://interstellar.com/${instance}/u/${user_name}`
`https://interstellar.com/${instance}/c/${community_name}/thread/${postId}`
`https://interstellar.com/${instance}/c/${community_name}/microblog/${postId}`

I also want to add routes for the different feed filters. e.g. `/all`, `/subscribed`, `/moderated` etc.
So yeah this PR should be for getting it working and then in another PR we can make sure the routes are consistent and comprehensive.

jwr1: I also think it would be nice to support a "default" instance, in case Interstellar is set up as an alternative frontend. For example, if I set up Interstellar running at i.kbin.earth, it would be nice if I could set kbin.earth to be the default server, and the URL path not include the domain if you are signed into kbin.earth. I think the path would look cleaner this way, and I could see this being useful if you want to view a link on the alternative frontend, just by changing the domain in your browser. So you might be able to do the following:

https://kbin.earth/u/jwr1 -> https://i.kbin.earth/u/jwr1
https://kbin.earth/m/steamdeck@sopuli.xyz/t/2360294/ -> https://i.kbin.earth/m/steamdeck@sopuli.xyz/t/2360294/ -> https://i.kbin.earth/c/steamdeck@sopuli.xyz/thread/2360294/ (we could set up a redirect in the router to make this once work.)

lib/src/widgets/redirect_listen.dart
olorin99: So it seems it isn't needed with the `useWebview` option set to true. However the app freezes. The authentication works but immediately after the webview closes the app freezes. Restarting the app opens with the account logged in so it does "technically" work just probably really annoying for users.

The error reported is
`embedder.cc (2575): 'FlutterEngineRemoveView' returned 'kInvalidArguments'. Remove view info was invalid. The implicit view cannot be removed.` and then a bunch of eglContext errors as flutter tries and fails to rebuild the opengl context.

lib/src/controller/router.dart
olorin99: That would be neat. Like being able to switch between old.reddit.com and reddit.com without changing the rest of the url.

lib/src/widgets/redirect_listen.dart
jwr1: Ok, we can leave it for now then.

olorin99:
Can confirm it works for me on linux.

lib/src/controller/database/database.dart
olorin99: @jwr1 Do we need this? Because it breaks the web build. Could probably write a conditional import to import either the native or web version but if it is only for debug builds it might not be worth it.

jwr1: No, we don't really need it, it was just recommended just for verification.

PR #359, opened 2026-01-28
Emoji picker for PieFed reactions and markdown editor

jwr1:
In this PR:

- PieFed emoji reactions are displayed above the post/comment action row (even in the feed).

- While signed in, clicking on a present emoji reaction will select the same emoji for you.
- Custom server specific emoji reactions are displayed.

- An emoji picker (used for PieFed reactions and MD editor), with support for searching emojis, and emoji groups with buttons to scroll to the section. Emoji data is pulled in from [emojibase](https://github.com/milesj/emojibase) on app build (with build_runner). Emoji searching uses a Trie data structure for super efficient string prefix lookup (based on emoji labels, tags, and emoticons).

In another PR:

- Custom emoji picker groups, including quick access emojis and PieFed server specific emojis.
- If the screen is too narrow for the current emoji picker, split emoji groups across two lines and use that as the picker width.

### DIFF
--- a/lib/src/widgets/content_item/action_buttons.dart
+++ b/lib/src/widgets/content_item/action_buttons.dart
@@ -1,8 +1,10 @@
 import 'package:flutter/material.dart';
+import 'package:interstellar/src/controller/controller.dart';
 import 'package:interstellar/src/widgets/emoji_picker/emoji_picker.dart';
 import 'package:interstellar/src/widgets/loading_button.dart';
 import 'package:material_symbols_icons/symbols.dart';
 import 'package:interstellar/src/utils/utils.dart';
+import 'package:provider/provider.dart';
 
 class ActionButtons extends StatelessWidget {
   const ActionButtons({
@@ -45,6 +47,8 @@ class ActionButtons extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final ac = context.watch<AppController>();
+
     final comments = Row(
       mainAxisSize: MainAxisSize.min,
       children: [
@@ -78,7 +82,7 @@ class ActionButtons extends StatelessWidget {
     final voting = Row(
       mainAxisSize: MainAxisSize.min,
       children: [
-        if (onEmojiReact != null)
+        if (onEmojiReact != null && !ac.profile.hideEmojiReactions)
           EmojiPicker(
             childBuilder: (onClick, focusNode) => IconButton(
               onPressed: onClick,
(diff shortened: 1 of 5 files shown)


lib/src/models/post.dart
```diff
@@ -102,6 +103,7 @@ abstract class PostModel with _$PostModel {
     required List<Tag> flairs,
     required PollModel? poll,
     required String? apId,
+    required List<EmojiReactionModel>? emojiReactions,
```
olorin99, line 106: Is there a benefit to having this nullable rather than just defaulting to an empty list? That way the `widget.emojiReactions != null && widget.emojiReactions!.isNotEmpty` check in content item could just be `widget.emojiReactions.isNotEmpty` since an empty list and a null would be treated the same anyway.

lib/src/widgets/content_item/content_item.dart
```diff
@@ -522,7 +526,44 @@ class _ContentItemState extends State<ContentItem> {
                         ),
                       ],
                     ),
-                  if (!profile.hideActionButtons)
+                  if (widget.emojiReactions != null &&
```
olorin99, line 529: Might be good to have a setting to disable emojis, since some people like them and some people really hate emojis with a passion.

lib/src/widgets/content_item/action_buttons.dart
```diff
@@ -74,6 +78,15 @@ class ActionButtons extends StatelessWidget {
     final voting = Row(
       mainAxisSize: MainAxisSize.min,
       children: [
+        if (onEmojiReact != null)
```
olorin99: Should also add an emoji picker to actions of the context menu. So that reacting is still available when the action buttons are hidden.

jwr1:
I think it would be a good idea to create an abstract menu system. One that displays full screen bottom sheet modals on small screens, but displays floating menus on large screens. I won't do that in this pr though.

lib/src/models/post.dart
jwr1: Where possible, I do like to have models use null to represent a missing feature, and an empty value to represent an existing feature but with no data. This means hopefully less platform checks in UI code. For example, when displaying the number of boosts on a button, instead of checking "is platform Mbin", we just check, "is post.boosts not null". In the case of emoji reactions, if Mbin later adds the ability, the idea is that we would only need to update the model and the API code, but not the UI code.

jwr1: Looking back at the code, it looks like my boosts example didn't hold, but you can at least see it with the `notificationControlStatus` property.

I'm sure you get the point though :)

jwr1: Unfortunately, it looks like PieFed returns `null` (instead of an empty list) when there are no reactions, so that kind of conflicts with how I was doing things.

olorin99: Yeah that makes sense.

PR #365, opened 2026-02-11
Add option to set default action for links.

### LINKED ISSUE #271: default option for link opening pop-up
### Describe the feature

add an option to decide a default option when opening links (copy link, open in browser, open in webview)

### Additional context

_No response_

olorin99:
Allows changing the default action when clicking on a link. 
Available actions include
- Share
- Copy
- Open in browser
- Open in webview
- Open secondary menu.

Closes #271

### DIFF
--- a/lib/src/widgets/open_webpage.dart
+++ b/lib/src/widgets/open_webpage.dart
@@ -28,6 +28,10 @@ enum LinkAction {
         shareUri(uri);
       case LinkAction.copy:
         await Clipboard.setData(ClipboardData(text: uri.toString()));
+        if (!context.mounted) return;
+        scaffoldMessengerKey.currentState?.showSnackBar(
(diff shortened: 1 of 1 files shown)


lib/src/widgets/open_webpage.dart
```diff
@@ -3,15 +3,47 @@ import 'dart:async';
...
+  openInBrowser,
+  openInWebview,
+  menu;
+
+  Future<void> call(BuildContext context, Uri uri) async {
+    switch (this) {
+      case LinkAction.share:
+        shareUri(uri);
+      case LinkAction.copy:
+        await Clipboard.setData(ClipboardData(text: uri.toString()));
```
jwr1, line 30: Could we have an indicator that the copy went through, just so it doesn't look like nothing happened when copy is the default? Maybe a 1 second snack bar or something similar?

PR #366, opened 2026-02-11
Moderate lock

olorin99:
Add moderator actions for locking posts on Mbin and Piefed and for locking comments on Piefed.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -375,6 +375,7 @@
   "settings_debug_clearReadPosts": "Clear read posts",
   "settings_debug_clearFeedCache": "Clear feed cache",
   "settings_debug_clearTags": "Clear user tags",
+  "settings_debug_clearOauth": "Clear Oauth identifiers",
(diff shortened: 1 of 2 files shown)


lib/src/models/post.dart
```diff
@@ -248,6 +251,7 @@ abstract class PostModel with _$PostModel {
       isPinned:
           lemmyPost['featured_community']! as bool ||
           lemmyPost['featured_local']! as bool,
+      isLocked: false,
```
jwr1:
```suggestion
      isLocked: lemmyPost['locked']! as bool ,
```

We could still use this to prevent people from trying to reply on a locked Lemmy post.

jwr1 (changes requested):
Just found those three things, looks great otherwise!

lib/src/models/post.dart
olorin99: Also added the lemmy api for locking a post. Moderator actions aren't enabled on lemmy yet but it's there for the future.

PR #371, opened 2026-02-18
Support UnifiedPush on Linux

jwr1:
The Flutter `unifiedpush` package added support for Linux a few months ago, so now we can finally support push notifications on Linux. I have the UnifiedPush storage using our database instead of the default shared preferences storage.

I have things mostly working, but unfortunately, there is an issue where once a notification is received in the background the first time, it will display correctly, but on subsequent notifications, it will no longer work. It seems like the app is staying open in the background and not closing properly after handling the notification.

### DIFF
--- a/lib/src/init_push_notifications.dart
+++ b/lib/src/init_push_notifications.dart
@@ -4,9 +4,7 @@ import 'dart:math';
 
 import 'package:auto_route/auto_route.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
-import 'package:http/http.dart' as http;
 import 'package:interstellar/src/controller/controller.dart';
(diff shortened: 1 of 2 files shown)


jwr1:
It sort of works, but unfortunately, the notification displays for about half a second before disappearing (while most notifications display for 10 seconds).

I'll ask in the UnifiedPush matrix just to see if they have any ideas.

jwr1:
After testing again, it seems like the issue is present, even if Interstellar is open in the foreground. That being, where the first notification will go through, but not subsequent.

jwr1:
The logs are coming in consistently every single time, but the actual displaying of the notification is extremely hit or miss. It seems like the app is fine staying open in the background for a little bit after a UP message is received because it still logs subsequent messages just fine. I can even see the "Initializing interstellar" line displayed only once with multiple "UnifiedPush message" lines afterwards.

So anyway, it is definitely an issue with `flutterLocalNotificationsPlugin.show`, but I'm really not sure what the problem would be.

PR #379, opened 2026-03-07
Add extra compact setting.

### LINKED ISSUE #378: Extra Compact / Headline Mode
### Describe the feature

As the headlines of posts are often cut off on Interstellar (even when the text is made much smaller), perhaps a display mode could be added that removes the post image, community icon, and user profile icon from posts to instead make room for a second row for the title?

As an illustration, here's Interstellar's compact mode versus the Compact mode on Mlem, Headline mode on Arctic, and Extra Compact mode on Blorp:

Image
Image
Image
Image

### Additional context

_No response_

olorin99:
Combine settings 'compactMode' and 'showPostsCards' into 'postMode' enum and add extra compact mode.
Post mode can be 
- card
- list
- compact
- extra compact

Extra compact mode doesn't show the image, title of the post isn't cutoff and doesn't show community/user icons.
Closes #378 

Extra compact vs compact
Screenshot from 2026-03-07 15-04-14-1

### DIFF
--- a/lib/src/widgets/content_item/content_item.dart
+++ b/lib/src/widgets/content_item/content_item.dart
@@ -918,38 +918,79 @@ class _ContentItemState extends State<ContentItem> {
                     ?contentTitle(context, null),
                     Row(
                       children: [
-                        DisplayName(
-                          widget.community!.name,
-                          onTap: () => context.router.push(
-                            CommunityRoute(
-                              communityName: widget.community!.name,
-                              communityId: widget.community!.id,
+                        Flexible(
+                          child: DisplayName(
+                            widget.community!.name,
+                            onTap: () => context.router.push(
+                              CommunityRoute(
+                                communityName: widget.community!.name,
+                                communityId: widget.community!.id,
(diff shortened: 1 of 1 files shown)


lib/src/controller/database/database.dart
```diff
@@ -481,6 +480,18 @@ class InterstellarDatabase extends _$InterstellarDatabase {
...
         from2To3: (m, schema) async {
           await m.addColumn(schema.profiles, schema.profiles.defaultLinkAction);
+          await m.addColumn(schema.profiles, schema.profiles.postMode);
+
+          await customUpdate(
+            "UPDATE profiles SET post_mode = 'compact' WHERE compact_mode = true",
+          );
+          await customUpdate(
+            "UPDATE profiles SET post_mode = 'list' WHERE show_posts_cards = false",
+          );
```
jwr1, line 490: Won't these override each other? It seems like they probably need to be swapped, since compact mode could only be enabled when show post cards is disabled.

PR #381, opened 2026-03-10
Piefed feeds.

olorin99:
Adds feed details for piefed feeds similar to the community details ui. Showing feed title, icon etc.
Adds feed menu with options for showing included communities, editing and saving locally.
Allows editing server feeds.


jwr1:
When deleting a feed, it just stays on the same screen as if nothing had happened. We probably want to navigate back from there instead.

jwr1 (approved):
Thanks!

PR #383, opened 2026-03-14
Add support for the modlog on piefed.

olorin99:
Is pretty much the same as lemmy api except some fields are nullable.


jwr1 (approved):
LGTM!

PR #384, opened 2026-03-23
Add support for Piefed events.

olorin99:
Adds support for viewing and creating events.

There are several constraints for creating an event and parts of the UI are highlighted when these constraints aren't met.
- Start date/time should be after the current date/time.
- End date/time should be after the start date/time.
- If online event is checked a link should be included.

## Viewing a event.
Screenshot from 2026-03-23 13-58-49

## Creating an event.
Screenshot from 2026-03-23 14-01-16

## With end date before start date.
Screenshot from 2026-03-23 14-12-55


lib/src/screens/feed/create_screen.dart
```diff
@@ -306,6 +311,97 @@ class _CreateScreenState extends State<CreateScreen> {
...
+                pickedDate = DateTime(
+                  pickedDate.year,
+                  pickedDate.month,
+                  pickedDate.day,
+                  pickedTime.hour,
+                  pickedTime.minute,
+                );
+                onSelect(pickedDate);
+              },
+              child: Text(date != null ? dateOnlyFormat(date) : 'Unset'),
```
jwr1: A localized string is needed here.

Actually, if both `_startDate` and `_endDate` are initialized to `Date.now()`, is `date` here even ever equal to `null`?

lib/l10n/app_en.arb
```diff
@@ -616,6 +616,16 @@
...
   },
+  "event": "Event",
+  "eventMode": "Event Mode",
+  "eventMode_free": "Free",
+  "eventMode_restricted": "Restricted",
+  "eventMode_external": "External",
+  "eventMode_invite": "Invite",
+  "eventOnline": "Online Event",
+  "eventAddEnd": "Add end date",
+  "eventError_start": "The start date/time is required to be after the current date/time.",
```
jwr1:
```suggestion
  "eventError_start": "The start date/time must be in the future.",
```
For clarity.

lib/l10n/app_en.arb
```diff
@@ -616,6 +616,16 @@
...
+  "eventError_end": "The end date/time is required to be before the start date/time.",
```
jwr1:
```suggestion
  "eventError_end": "The end date/time must be after the start.",
```

lib/src/api/threads.dart
```diff
@@ -534,6 +532,59 @@ class APIThreads {
...
+    bool online = true,
+    String feeCurrency = 'USD',
+    int fee = 0,
+  }) async {
+    switch (client.software) {
+      case ServerSoftware.mbin:
+        throw UnimplementedError('Polls are unsupported on mbin');
+
+      case ServerSoftware.lemmy:
+        throw UnimplementedError('Polls are unsupported on lemmy');
```
jwr1:
```suggestion
      case ServerSoftware.mbin:
        throw UnimplementedError('Events are unsupported on mbin');

      case ServerSoftware.lemmy:
        throw UnimplementedError('Events are unsupported on lemmy');
```

lib/src/screens/feed/create_screen.dart
```diff
@@ -306,6 +311,97 @@ class _CreateScreenState extends State<CreateScreen> {
...
+        child: Row(
+          mainAxisAlignment: MainAxisAlignment.spaceBetween,
+          children: [
+            TextButton(
+              onPressed: () async {
+                var pickedDate = await showDatePicker(
+                  context: context,
+                  firstDate: DateTime.now(),
+                  lastDate: DateTime.now().add(const Duration(days: 365 * 10)),
+                );
```
jwr1: If we pass in `initialDate`, then when editing the date, it will be preselected with whatever the previous value was. That should make editing the date more than once easier.

lib/src/screens/feed/create_screen.dart
```diff
@@ -306,6 +311,97 @@ class _CreateScreenState extends State<CreateScreen> {
...
+                if (pickedDate == null) return;
+                onSelect(pickedDate);
+
+                if (!context.mounted) return;
+
+                final pickedTime = await showTimePicker(
+                  context: context,
+                  initialTime: const TimeOfDay(hour: 0, minute: 0),
+                );
```
jwr1: Same for initialTime and having it preselected to the current value.

lib/src/utils/utils.dart
```diff
@@ -61,12 +62,12 @@ String dateDiffFormat(DateTime input) {
...
-  if (hours > 0) {
+  if (hours.abs() > 0) {
     return '${hours}h';
   }
 
   final minutes = difference.inMinutes;
-  if (minutes > 0) {
+  if (minutes.abs() > 0) {
     return '${minutes}m';
   }
```
jwr1, line 73: What are the `abs()` here for? Are we expecting `end` to ever be before `start`?

jwr1 (changes requested):
Sorry for taking so long to review; I've been busy this semester. Anyway, it looks pretty good!

lib/src/screens/feed/create_screen.dart
olorin99: End date used to be nullable. Forgot to fix this bit when I switched it to being required.

lib/src/utils/utils.dart
olorin99: Not for event but it fixed another date which I have now forgotten what it was. I think it was a post published/edited date that was weird.

jwr1 (approved):
LGTM!!!

PR #386, opened 2026-04-03
Open activitypub urls within app.

### LINKED ISSUE #393: Support deep links from fediverse servers
### Describe the feature

Feature request from Matrix:

> I was wondering if there were any plans to add instances to the android "open by default" links? Currently Interstellar doesn't support the opening of any links on android, and would need to be explicitly added to the manifest.

### Additional context

We would need a reliable method of fetching a list of all Mbin, Lemmy, and PieFed servers at build time. Deep links are not customizable by the user, so they would not be able to add their own server if they wanted to.

olorin99:
Open activitypub links in the app instead of opening in a browser.

~~On app start/ account login fetches list of federated instances and caches locally. When a link is clicked the host is matched against this list to determine whether it is an activitypub server or not. If it is then attempt to resolve the url and open the appropriate post/comment route.~~

To resolve the url `/resolve_object` is used on lemmy and piefed and `/search` is used on mbin. 

On mbin it can be kind of slow and doesn't always resolve urls properly. Testing these urls https://kbin.earth/m/testing/t/2536658 about half worked properly on mbin while piefed handled all of them.


jwr1:
I tried running the app locally for a quick test on my device; however, it seemed like the app was hanging on something when the home feed tried loading, as the loading indicator never went away.

I'll loop back to this pr to review it more after I take a look at the others.

olorin99:
I've updated this to use a static list of known instances rather than fetching a list at runtime which should be more stable. I pretty much copied the script for fetching a list of instances from Thunder so thanks to them.

I'm also using this new list for handling deeplinks. Deeplinks can either be enabled for Interstellar specifically or by going through LemmyRedirect.

olorin99:
Since there have been some additions to Mbin recently that are related to this PR I think we should wait til Mbin's next full release before merging.

olorin99:
The new search endpoint for Mbin seems to work well. I've also added support for deeplinks for popular mastodon instances.

PR #388, opened 2026-04-09
Lemmy moderation

olorin99:
Adds endpoint for moderation actions on lemmy.

Made some changes to the reports list UI. I found the accept/reject labels for mbin moderation a little confusing. It not immediately clear what accepting or rejecting a report does. So I've changed it to be more like most of the lemmy alt frontends by replacing the accept/reject buttons with delete and resolve buttons. 
Screenshot from 2026-04-09 20-25-42


jwr1 (changes requested):
You can't unresolve a report on Mbin though, can you (the button is still clickable)? Same with the restore button as well for Mbin.

jwr1:
Also, if we are changing the terminology for the buttons, should we do so as well for the displayed report status and also the reports filter?

olorin99:
> You can't unresolve a report on Mbin though, can you (the button is still clickable)? Same with the restore button as well for Mbin.

Yeah its emulating the lemmy way of resolving the report being separate from deleting the post. The delete/restore button sets the state the post should be in then resolve/unresolve submits that state.

I'm thinking now that it probably would be cleaner to go the other way around and have lemmy adapt to the current UI and just change the labels. 

Accept is deleting the post and reject is keeping it right? I understand the logic of why accepting the report deletes the post and rejecting the report keeps the post but my brain keeps twisting it around and confusing me.

olorin99:
> Also, if we are changing the terminology for the buttons, should we do so as well for the displayed report status and also the reports filter?

Yeah I'll make this change too.

jwr1:
> > You can't unresolve a report on Mbin though, can you (the button is still clickable)? Same with the restore button as well for Mbin.
> 
> Yeah its emulating the lemmy way of resolving the report being separate from deleting the post. The delete/restore button sets the state the post should be in then resolve/unresolve submits that state.
> 
> I'm thinking now that it probably would be cleaner to go the other way around and have lemmy adapt to the current UI and just change the labels.
> 
> Accept is deleting the post and reject is keeping it right? I understand the logic of why accepting the report deletes the post and rejecting the report keeps the post but my brain keeps twisting it around and confusing me.

I have to admit, I did find the new labels initially confusing (the way Lemmy has it), but that could just be that I'm just so used to clicking on buttons "Approve" or "Reject" through moderating kbin.earth.

I'll let you use your judgment though, and we can go with whichever terminology you decide to use.

jwr1 (approved):

PR #390, opened 2026-04-11
Add filter for feeds to switch between all feeds and personal feeds.

olorin99:


lib/src/screens/explore/explore_screen.dart
jwr1, line 1: If the user is logged out, we shouldn't allow filtering by personal feeds.

jwr1, line 1: Since personal feeds are not shown when "All results" is selected, we might want to use different terminology. Maybe "Public" and "Private" feeds, unless PieFed specifically refers to non-public feeds as "personal".

olorin99: Problem is that its not a clean split between public/private. The personal filter shows both the users public and private feeds. Piefed just refers to "my feeds". I thought personal would be good as its just one word.

Maybe "Public" and "Personal" as the filters.

jwr1: Ah, I see what you mean. Yeah, Public and Personal seems like it could work.

jwr1 (approved):

PR #396, opened 2026-04-29
Enable using domains in feeds on mbin.

olorin99:
Uses `/api/domains?=search` to get a domain by its name instead of by id.


jwr1 (approved):
Thanks!

PR #399, opened 2026-05-12
Download settings

### LINKED ISSUE #352: Allow downloading of settings (profiles)
### Description

From a Google Play review:

> Recommend. Though, I would like to see a feature implemented: sometimes the app gets into an error where it physically will not refresh the feed and instead just says there's an error and can't refresh, with the only fix being to reinstall the app. since reinstalling and signing back in completely erases your settings in the app such as blocked word lists, I'd like actually if we could download our settings from the app then apply them to a profile. newpipe does something similar.

olorin99:
Allows downloading and importing profiles, filter lists and feeds as JSON files.
This is only for downloading individual items not for making a backup of entire app state but the database import/export is more appropriate for that.

Closes #352

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -899,5 +899,8 @@
         }
     }
   },
-  "invalidFile": "Invalid file"
+  "invalidFile": "Invalid file",
+  "downloadFeed": "Download feed",
+  "asJSON": "As JSON",
+  "asOPML": "As OPML"
 }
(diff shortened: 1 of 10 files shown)


olorin99:
I also added support for exporting feeds as OPML files to be used in RSS readers. Doesn't support all feed sources as not all platforms have corresponding RSS feeds for each source. Can export community and user RSS for Lemmy and Mbin but couldn't find RSS url for Piefed users. Closes #401

PR #409, opened 2026-07-09
Combined API

olorin99:
Switch to mbin combined API rather than merging threads/microblogs client side. There's a option for including boosts in the results but I was getting timeouts when it was enabled so left it off by default.


jwr1 (approved):
Thank you, it works well!

PR #420, opened 2026-09-04
Validate Mbin community names when creating a community

neo22neo:
## Problem

When creating a community on an Mbin server, the form only disabled the submit button on an empty name. Mbin rejects any magazine name outside `/^[a-zA-Z0-9_]{2,25}$/` (`RegPatterns::MAGAZINE_NAME`), so names with spaces, hyphens, accents, etc. failed server-side with no guidance for the user.

## Changes

- **`lib/src/utils/mbin_community_name.dart`** (new) — pure, testable helpers:
  - `isValidMbinCommunityName` — the `^[a-zA-Z0-9_]{2,25}$` check.
  - `mbinCommunityNameIssue` — why a name is invalid (`invalidCharacters` / `tooShort` / `tooLong`); empty is treated as "not entered yet".
  - `suggestMbinCommunityName` — best-effort sanitized name (unsupported runs → `_`, collapsed, trimmed, truncated to 25), or `null` when nothing usable can be salvaged.
- **`lib/src/widgets/text_editor.dart`** — shared `TextEditor` gains `helperText` / `errorText` passthrough.
- **`lib/src/screens/explore/community_owner_panel.dart`** — in `CommunityOwnerPanelGeneral` (creation path only), when the active server is Mbin:
  - show the naming rule as helper text,
  - show an inline error while the name is invalid,
  - offer a one-tap **Use “…”** action to apply the suggested name,
  - keep the submit button disabled until the name is valid.
  - Lemmy / PieFed behaviour is unchanged (still `maxLength: 25`, no extra validation).
- **`lib/l10n/app_en.arb`** — new keys: `community_nameMbinHelp`, `community_nameInvalidCharacters`, `community_nameTooShort`, `community_nameTooLong`, `community_nameUseSuggestion`.
- **`pubspec.yaml` / `test/`** — add `flutter_test` dev dependency and focused unit tests for the new helpers.

## Notes

- The regex matches upstream Mbin (`MbinOrg/mbin` `src/Utils/RegPatterns.php`).
- I don't have a Flutter toolchain in this environment, so `dart format` / `flutter test` / `flutter gen-l10n` were not run locally — please double-check CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/src/screens/explore/community_owner_panel.dart
```diff
@@ -112,23 +114,75 @@ class _CommunityOwnerPanelGeneralState
...
   }
 
+  String? _mbinNameError(BuildContext context, MbinCommunityNameIssue? issue) {
+    switch (issue) {
+      case MbinCommunityNameIssue.invalidCharacters:
+        return l(context).community_nameInvalidCharacters;
+      case MbinCommunityNameIssue.tooShort:
+        return l(context).community_nameTooShort(mbinCommunityNameMinLength);
+      case MbinCommunityNameIssue.tooLong:
+        return l(context).community_nameTooLong(mbinCommunityNameMaxLength);
```
jwr1, line 124: Instead of enforcing the max community length as an error, why don't you use the `maxLength` field on the `TextEditor` widget? It would display the current and max length of the text as the user is typing, and prevent the user from typing over that length.

lib/src/utils/mbin_community_name.dart
```diff
@@ -0,0 +1,61 @@
...
+  var suggestion = name
+      .replaceAll(_mbinCommunityNameInvalidChars, '_')
+      .replaceAll(RegExp(r'_+'), '_')
+      .replaceAll(RegExp(r'^_+|_+$'), '');
+
+  if (suggestion.length > mbinCommunityNameMaxLength) {
+    suggestion = suggestion
+        .substring(0, mbinCommunityNameMaxLength)
+        .replaceAll(RegExp(r'_+$'), '');
+  }
```
jwr1, line 55: Would it work to move the `.replaceAll(RegExp(r'^_+|_+$'), '')` portion from line 49 to after the if statement, that way we could remove the `.replaceAll(RegExp(r'_+$'), '')` portion from line 54 that's in the if statement?

lib/src/utils/mbin_community_name.dart
```diff
@@ -0,0 +1,61 @@
...
+  return null;
+}
+
+enum MbinCommunityNameIssue { invalidCharacters, tooShort, tooLong }
+
+/// A best-effort valid name derived from [name], or `null` when nothing
+/// usable can be salvaged (e.g. the input has no letters/digits at all) or
+/// when [name] is already valid.
+String? suggestMbinCommunityName(String name) {
```
jwr1, line 40: One addition that would be nice for this function is to also make the suggested name all lowercase, as that seems to be the general convention for community names.

jwr1 (changes requested):
Hello @neo22neo, thank you for submitting your first PR here! I especially like the idea of suggesting a community name based on the title someone has set already. I haven't checked out the code locally yet, but upon first review of the code, I do have a few things to note.

PR #421, opened 2026-09-05
Render server validation errors as form errors

neo22neo:
> **Stacked on #420.** This branch is based on `fix/mbin-community-name-validation`,
> so until #420 merges this PR shows that commit and its files too. The only work
> owned by this PR is the last commit, `Render server validation errors as form
> errors` (`lib/src/api/client.dart` + a small delta in
> `lib/src/screens/explore/community_owner_panel.dart`). Once #420 is merged I'll
> rebase so the diff here collapses to just that. Please merge #420 first.

## Problem

#420 stops *predictable* bad names — anything outside Mbin's
`^[a-zA-Z0-9_]{2,25}$` — before the request is ever sent. But the server still
rejects names for reasons a client cannot know in advance: already taken,
reserved, or blocked by instance policy. Those come back as `400` with an
RFC 7807-style body:

```json
{
  "type": "https://tools.ietf.org/html/rfc2616#section-10",
  "title": "An error occurred",
  "status": 400,
  "detail": "..."
}
```

`ServerClient.checkResponseSuccess` embedded that entire JSON body into an
`http.ClientException` message, and the create screen had no `try`/`catch`, so
it reached the global handler in `main.dart` and rendered as a snackbar full of
raw JSON. The useful `detail` was buried inside it.

So: #420 handles malformed names before submit; **this PR makes the server's own
validation responses render as form errors instead of raw
`ClientException`/JSON**. It's the backstop, not the primary fix.

## Changes

**`lib/src/api/client.dart`**
- New `ServerErrorException` carrying `statusCode` / `uri` / `title` / `detail` /
  `rawBody`, with `toString()` returning `detail ?? title ?? <generic>`.
- `checkResponseSuccess` now tries to parse an error body as a JSON object with
  a string `title`/`detail` and throws `ServerErrorException` when it can.
  Bodies that aren't structured problem JSON — Lemmy/PieFed `{"error": ...}`,
  HTML, empty — still throw `http.ClientException` exactly as before. No call
  site in the repo matches `ClientException` by type, so this is backwards
  compatible and incidentally means every generic `catch (e)` now shows a clean
  message instead of a JSON blob.

**`lib/src/screens/explore/community_owner_panel.dart`**
- `create()` is wrapped in `try / on ServerErrorException`: a `400` carrying a
  `detail` is shown on the Name field through the `errorText` plumbing added in
  #420, and swallowed so no snackbar fires. Anything else is rethrown to the
  existing generic handling.
- The message is stored alongside the name it was returned for and only shown
  while the field still holds that name, so it clears itself on any edit —
  including #420's suggestion button, which sets the controller text directly
  and so never fires `onChanged`.
- The create/edit ternary became an `if`/`else` so only the create call is
  wrapped; the edit path is behaviourally unchanged.

This PR deliberately contains **no** validation, helper-text, suggestion, or
`TextEditor` changes — all of that is #420's.

## Testing

No Flutter toolchain in this environment, so `dart format` / `flutter test` /
the app itself were not run locally — CI is the first real check. Verified by
hand: the sample body above parses with `detail` as a `String`, and no line in
the changed files exceeds 80 columns.

Manual test plan (Mbin account), after #420:
1. Name that #420 already rejects → inline error, submit blocked, no request.
2. Well-formed name that the server rejects anyway (e.g. one already taken) →
   the server's `detail` appears under the Name field, no snackbar.
3. Edit the name → the server error clears; tap #420's suggestion button → it
   also clears.
4. Force an unrelated failure (offline, 500) → still the generic snackbar.
5. Lemmy/PieFed creation and the Mbin community *edit* screen unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015icyWnYYmR7J8oSHocta6U


neo22neo:
@jwr1 could you take a look when you have a chance? This turns the Mbin 400 name-validation response into an inline field error instead of the raw-JSON snackbar. I couldn't run `dart format` locally (no Flutter toolchain here), so if the analysis job flags whitespace I'll push a formatting-only fixup.

neo22neo:
Restructured this PR: it was independently duplicating #420's client-side validation and `TextEditor` changes. It's now stacked on `fix/mbin-community-name-validation` and reduced to only the server-error handling — `ServerErrorException` in `api/client.dart` plus a small delta in the create screen that routes a 400's `detail` through the `errorText` plumbing #420 adds. All duplicated validation/helper/TextEditor code has been removed. Please merge #420 first; I'll rebase afterwards so the diff here collapses to just the last commit.

jwr1:
With the additional client side validations in #420, I'm not sure how much this PR is needed. I do think it would be good to have more human friendly error message pop-ups from the snack bar. However, if we wanted to switch away from the snack bar to inline error messages, that would take a huge undertaking in order to rework all the other places in the UI where errors could occur.

